PR #2, opened 2017-02-17
Refine the job runner.

michaelfinch:
PRs after this will be a lot smaller. Probably better to just look at the entire source instead of the diff


daniel-nichter:
Awesome 👍

PR #3, opened 2017-02-22
Repurpose cache package as a more general db package

michaelfinch:
Reason for this change: there already existed a cache.Cacher interface that allowed storing keys/values in a memory cache, but I realized the interface could be used more generically to store key/values anywhere (ex: in redis), so the interface was renamed to db.Driver. The memory implementation of the interface wasn't changed.

chain.Repo can now create a repo that is backed by a memory db (used to just be a fake stub), and support for redis will be added there once the redis driver is implemented.


cuzelac:
This feels like a premature optimization to me, as we haven't many backends yet.

Since the `db.Driver` interface apes the `chain.Repo` interface so closely, I expect that the two interfaces will have to grow in lock-step as features are added.

This driver interface is very specific to key/value stores, so its name should be changed to reflect that, maybe `db.KVDriver` or something.

PR #8, opened 2017-02-28
Chain repo implementation backed by Redis

cuzelac:
This does not yet use the [`db.Driver` interface](https://github.com/square/spincycle/blob/master/job-runner/db/datastore.go#L21).  I'll work to get that going in a future version.


job-runner/chain/redis_repo.go
```diff
@@ -0,0 +1,185 @@
...
+	"sync"
+	"time"
+
+	"github.com/garyburd/redigo/redis"
+)
+
+// ChainKey is the keyspace for serialized Chains
+const CHAIN_KEY = "ChainById"
+
+var (
```
daniel-nichter: These vars should be in `repo.go`, i.e. one source of truth wrt repo errors so the caller (whatever code is using the repo) can remain oblivious to the implementation (Redis repo, memory repo, etc. -- doesn't matter, they all return the same set of errors).

job-runner/chain/redis_repo.go
```diff
@@ -0,0 +1,185 @@
...
+	ErrKeyExists           = errors.New("key already exists in redis")
+	ErrKeyNotFound         = errors.New("key not found in redis")
+	ErrMultipleKeysDeleted = errors.New("multiple keys deleted from redis")
+)
+
+// A RedisConnectionPool is used for pooling redis connections.
+type RedisConnectionPool interface {
```
daniel-nichter: So far I don't think this needs to be exported/public. Would the creator (probably just main.go) need to provide this? That seems too much to ask callers. More likely, caller will create a Redis repo with an arg like `maxConnections` or `poolSize` or something, then leave it to the Redis repo itself to handle the internal details. (Looking below, I see `RedisRepoConfig` could be used for this.)

job-runner/chain/redis_repo.go
```diff
@@ -0,0 +1,185 @@
...
+
+type RedisRepo struct {
+	ConnectionPool RedisConnectionPool // connection pool
+	Conf           RedisRepoConfig     // config this Repo was built with
+}
+
+// NewRedisRepo builds a new Repo backed by redis
+func NewRedisRepo(c RedisRepoConfig) (*RedisRepo, error) {
+	// build connection pool
+	addr := c.Server + ":" + strconv.FormatUint(uint64(c.Port), 10)
```
daniel-nichter: `fmt.Sprintf` might be easier :-)

job-runner/main.go
```diff
@@ -14,9 +14,21 @@ import (
 )
 
 func main() {
+	// Make a chain repo.
+	chainRepo := chain.NewMemoryRepo()
```
daniel-nichter: Not using Redis repo? Or, how do we let user configure this? As OSS we can't assume user has/wants Redis, but then there should be some switch between memory-backed repo and Redis-backed repo if Redis is available.

job-runner/chain/redis_repo.go
michaelfinch: Right now this is only used for mocking out a pool in tests. redigo doesn't define this interface, so we have to

job-runner/chain/redis_repo.go
michaelfinch: ack

job-runner/main.go
michaelfinch: At some point you will be able to set all of this easily in a config file, but we don't have such a thing in place yet so if you want to change your repo type away from the memory repo right now you have to edit main.go (see comments below this line)

daniel-nichter: For the sake of this PR, we can punt this and do another PR later where we introduce a config file so we and users can specify a Redis `addr:port` and if specified then use Redis, else fallback to memory.

PR #16, opened 2017-03-22
Add an http client for the job-runner

michaelfinch:


job-runner/client/http.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func (c *jrClient) NewJobChain(jobChain proto.JobChain) error {
+	// POST /api/v1/job-chains
+	httpMethod := "POST"
+	url := c.baseUrl + "/api/v1/job-chains"
+	successStatusResp := 200
+
+	// Create the payload.
+	payloadBytes, err := json.Marshal(jobChain)
```
daniel-nichter: nit: "bytes" is commonly used (because, you know, Go loves shortest possible var names)

job-runner/client/http.go
```diff
@@ -0,0 +1,169 @@
...
+		return err
+	}
+
+	// Send the request.
+	resp, body, err := c.do(req)
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode != successStatusResp {
```
daniel-nichter: Better to use Go's `http.StatusOK` than defining our own `successStateResp` values.

job-runner/client/http.go
```diff
@@ -0,0 +1,169 @@
...
+			resp.StatusCode, string(body))
+	}
+
+	return nil
+}
+
+func (c *jrClient) StartRequest(requestId uint) error {
+	// PUT /api/v1/job-chains/${requestId}/start
+	httpMethod := "PUT"
+	url := c.baseUrl + "/api/v1/job-chains/" + strconv.FormatUint(uint64(requestId), 10) + "/start"
```
daniel-nichter: `url := fmt.Sprintf(baseUrl + "/api/v1/job-chains/%d/start", requestId)`

job-runner/client/http.go
```diff
@@ -0,0 +1,169 @@
...
+
+	if resp.StatusCode != successStatusResp {
+		return fmt.Errorf("unsuccessful status code: %d (response body: %s)",
+			resp.StatusCode, string(body))
+	}
+
+	return nil
+}
+
+func (c *jrClient) StopRequest(requestId uint) error {
```
daniel-nichter: This func and `NewJobChain` and `StartRequest` are nearly identical, they just use a different API endpoint. Maybe it's possible to factor out the common code like:
```go
StopRequest(...) {
  return post(<url specific to this>)
}
StartRequest(...) {
  return post(<url specific to this>)
}

post(...) {
  // the common code
}
```

michaelfinch:
addressed all feedback

PR #20, opened 2017-05-18
Initial Request Manager PR (and related Job Runner changes).

michaelfinch:


michaelfinch:
This PR has gotten excessively large - that's my bad, sorry about that. It started off with me just working on the request manager, but that required some changes to the job runner, which made me want to make further changes to the job runner. Anyway, I've got all the tests passing and it's basically done now (although I'm sure I'll be pushing very minor changes unti it gets reviewed). Here's a summary of what's in this pr:
- Created the request manager with an omega stub. Right now the request manager is an api that talks to a database and also with the job-runner
- Updated the job-runner to communicate back with the request manager when jobs/chains finish
- Removed the router package and changed the APIs to be built on an echo web server (https://echo.labstack.com/)
- Removed the kv package; now using concurrent-map (github.com/orcaman/concurrent-map)
- Created a config package so that all the settings in the main.go files can be changed purely from config
- Removed the job-runner/runner package since, as of now, it was providing almost 0 functionality and it was easy to get rid of

daniel-nichter:
Feedback on schema:

  * How about `s/requestor/user`? I think "user" is more standard, simpler, and more directly reflects the intent of logging _who_ (i.e. a person not a machine) made the request
  * Cols like `job_id` need to be `varchar(100) character set 'latin1'` to avoid multi-byte chars which I presume shouldn't be present in IDs
  * Why 175 for `job_id` and `type`?
  * Why 64 for `requests.type`?  That's a trick question :-) That's probably just a random number I chose in the design doc
  * We need to think more about `job_log` indexing because this table will grow rapidly
  * What's `lock_version`?
  * I prefer `s/id/request_id/` to keep pk cols uniformly named across all tables. Simpler to have 1 name for the same logical col. Also makes `JOIN ON <col>` possible.

Will provide feedback on code, too...

PR #22, opened 2017-06-13
omega

BrianIp:


michaelfinch:
A nice improvement for this PR would be better error reporting. Playing around with this package I often get the error "Job Args are missing" when creating a graph, which is really hard to track down without context.

PR #26, opened 2017-08-03
Update the grapher Node struct to store the Args that the node was cr…

michaelfinch:
…eated with, and also update the Job proto (which is part of the Job Chain) to store this as well

Goal: be able to tell what args a job was created with. For example, imagine you're looking at a job chain in the UI...if there are two jobs of the same type, how do you tell the difference between the two? By their jobArgs.

Solution: store a job's JobArgs directly in the proto.JobChain struct

Other considered solution: add a `Args() map[string]interface{}` method to the `Job` interface. that way, if you have a job, you can always run job.Args() to see what jobArgs it was created with. The downside with this solution is that spincycle has to recreate a job every time you want to see its args (so it would have to recreate lots of jobs every time you reload the web page displaying a job chain)

### DIFF
--- a/proto/s2s.go
+++ b/proto/s2s.go
@@ -15,6 +15,7 @@ type Job struct {
 	Type  string                 `json:"type"`  // user-specific job type
 	Bytes []byte                 `json:"bytes"` // return value of Job.Serialize method
 	State byte                   `json:"state"` // STATE_* const
+	Args  map[string]interface{} `json:"args"`  // the jobArgs a job was created with
 	Data  map[string]interface{} `json:"data"`  // job-specific data during Job.Run
 }
 
@@ -22,8 +23,8 @@ type Job struct {
 // Job chains are identified by RequestId, which must be globally unique.
 type JobChain struct {
 	RequestId     string              `json:"requestId"`     // unique identifier for the chain
-	Jobs          map[string]Job      `json:"jobs"`          // Job.Name => job
-	AdjacencyList map[string][]string `json:"adjacencyList"` // Job.Name => next []Job.Name
+	Jobs          map[string]Job      `json:"jobs"`          // Job.Id => job
+	AdjacencyList map[string][]string `json:"adjacencyList"` // Job.Id => next []Job.Id
 	State         byte                `json:"state"`         // STATE_* const
 }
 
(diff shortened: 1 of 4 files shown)


proto/s2s.go
```diff
@@ -15,6 +15,7 @@ type Job struct {
 	Type  string                 `json:"type"`  // user-specific job type
 	Bytes []byte                 `json:"bytes"` // return value of Job.Serialize method
 	State byte                   `json:"state"` // STATE_* const
+	Args  map[string]interface{} `josn:args"`   // the jobArgs a job was created with
```
felixplajer: typo "josn"

proto/s2s.go
```diff
@@ -15,6 +15,7 @@ type Job struct {
 	Type  string                 `json:"type"`  // user-specific job type
 	Bytes []byte                 `json:"bytes"` // return value of Job.Serialize method
 	State byte                   `json:"state"` // STATE_* const
+	Args  map[string]interface{} `json:args"`   // the jobArgs a job was created with
```
felixplajer: 😛  another typo - missing the first quote in args

request-manager/grapher/grapher.go
```diff
@@ -692,6 +695,12 @@ func (o *Grapher) newNoopNode(name string, nodeArgs map[string]interface{}) (*No
 
 // newNode creates a node for the given job j
 func (o *Grapher) newNode(j *NodeSpec, nodeArgs map[string]interface{}) (*Node, error) {
+	// Make a copy of the nodeArgs before this node gets created and potentially
+	// adds additional keys to the nodeArgs.
+	originalArgs := map[string]interface{}{}
+	for k, v := range nodeArgs {
+		originalArgs[k] = v
+	}
```
felixplajer: If one of the job args is something like a map (so only the reference is copied into originalArgs), and the job modifies it during Create, this will end up modifying the arg in job.Args too. Is this ok and/or intentional?

michaelfinch: It should not be possible for values in the nodeArgs map to change, but you bring up a good point so I will add a comment explaining that. Good eye Felix! There are a few other spots in the grapher where we copy args maps as well

proto/s2s.go
michaelfinch: 😮

PR #27, opened 2017-08-17
Various tweaks to job retry code

daniel-nichter:
* Standardize on term "(re)try" (singular and plural); remove "attempt"
* Rename proto.Job.RetriesAllowed -> .Retries and .RetryDelay to .RetryWait
* Rename proto.JobLog.Attempt -> .Try
* Rename job_log.attempt col -> .try
* Change job_log.try col default from 1 -> 0 (to detect bad rows)
* Remove extra (unused?) indexes on job_log table
* Change RetryWait from seconds to milliseconds
* Make var names more consistent and/or clear (finalState, pJob, runnableJob)
* Factor pJob fields out of Runner.Run() (bake into instance in NewRunner())

### DIFF
--- a/job-runner/chain/traverser.go
+++ b/job-runner/chain/traverser.go
@@ -295,22 +295,24 @@ func (t *traverser) Status() (proto.JobChainStatus, error) {
 // and attaches it to the job. When it's done, it sends the job out through the
 // doneJobChan.
 func (t *traverser) runJobs() {
-	for runJ := range t.runJobChan {
-		go func(j proto.Job) {
+	for runnableJob := range t.runJobChan {
+		go func(pJob proto.Job) {
+			// Always return the job when done, else the traverser will block.
+			defer func() { t.doneJobChan <- pJob }()
+
 			// Make a job runner. If an error is encountered, set the
 			// state of the job to FAIL and create a JL with the error.
-			runner, err := t.rf.Make(j, t.chain.RequestId())
+			runner, err := t.rf.Make(pJob, t.chain.RequestId())
 			if err != nil {
-				j.State = proto.STATE_FAIL
-				t.sendJL(j, err)   // need to send a JL to the RM so that it knows this job failed
-				t.doneJobChan <- j // send the job to the doneJobChan
+				pJob.State = proto.STATE_FAIL
+				t.sendJL(pJob, err) // need to send a JL to the RM so that it knows this job failed
 				return
 			}
 
 			// Add the runner to the repo. Runners in the repo are used
 			// by the Status and Stop methods on the traverser.
-			t.runnerRepo.Set(j.Id, runner)
-			defer t.runnerRepo.Remove(j.Id)
+			t.runnerRepo.Set(pJob.Id, runner)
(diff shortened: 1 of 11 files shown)


request-manager/grapher/grapher.go
```diff
@@ -50,12 +50,12 @@ type Graph struct {
...
-	Prev       map[string]*Node       // in edges ( node name -> Node )
-	Args       map[string]interface{} // the args the node was created with
-	Retries    int                    // the number of times to retry a node
-	RetryDelay int                    // the time, in seconds, to sleep between retries
+	Datum     Payload                // Data stored at this Node
+	Next      map[string]*Node       // out edges ( node name -> Node )
+	Prev      map[string]*Node       // in edges ( node name -> Node )
+	Args      map[string]interface{} // the args the node was created with
+	Retries   uint                   // the number of times to retry a node
+	RetryWait uint                   // the time, in seconds, to sleep between retries
```
felixplajer: comment should say milliseconds, not seconds.

job-runner/chain/traverser.go
```diff
@@ -295,22 +295,24 @@ func (t *traverser) Status() (proto.JobChainStatus, error) {
 // and attaches it to the job. When it's done, it sends the job out through the
 // doneJobChan.
 func (t *traverser) runJobs() {
-	for runJ := range t.runJobChan {
-		go func(j proto.Job) {
+	for runnableJob := range t.runJobChan {
+		go func(pJob proto.Job) {
+			// Always return the job when done, else the traverser will block.
+			defer func() { t.doneJobChan <- pJob }()
```
alyssa: is replacing the adding the job to the done job chain throughout the function with the defer here just a code clean up?

job-runner/runner/runner.go
```diff
@@ -35,12 +35,15 @@ type Runner interface {
...
-	pJob    proto.Job // struct with info about the job
 	realJob job.Job   // the actual job interface to run
 	reqId   string    // the request id the job belongs to
 	rmc     rm.Client // client used to send JLs to the RM
 	// --
-	stopChan chan struct{} // used on Stop
+	jobId     string
+	jobType   string
+	maxTries  uint
+	retryWait time.Duration
```
alyssa: It seems like `jobId`, `jobType`, `maxTries`, and `retryWait` are all in `proto.Job`. Are we listing them out here rather than use the `proto.Job` because we don't need to reference all of the fields in the struct or because we want to capture the data at this point in time?

job-runner/chain/traverser.go
daniel-nichter: Cleanup and improvement. Previously, iirc, there were returns on error before `pJob` was sent back, which means it never got sent back. But `pJob` should always be sent back regardless, so `defer` at the start is the solution.

job-runner/runner/runner.go
daniel-nichter: Replied elsewhere, but in short: it decouples proto.Job from runner internals, allowing us to change proto.Job and then have only one place to update runner, rather than having to update runner's direct use of proto.Job throughout its internals.

PR #29, opened 2017-08-30
Add spinc cli

daniel-nichter:


request-manager/status/manager.go
```diff
@@ -0,0 +1,127 @@
...
+	for _, r := range running {
+		if seen[r.RequestId] {
+			continue
+		}
+		ids = append(ids, r.RequestId)
+		seen[r.RequestId] = true
+	}
+
+	q := "SELECT request_id, type, state, user, created_at, started_at, finished_at, total_jobs, finished_jobs" +
+		" FROM requests WHERE request_id IN (" + db.IN(ids) + ")"
```
michaelfinch: for security, should we use query params here instead of injecting ids directly into the sql?

PR #30, opened 2017-08-30
Ensure response body closes when there's an error.

alyssa:


alyssa:
Nevermind. I assumed we always wanted to close the response body, but the http docs says the current way is the correct way!

PR #33, opened 2017-09-14
Make the RM more modular

michaelfinch:


config/config.go
```diff
@@ -92,6 +92,10 @@ type SQLDb struct {
 
 	// The TLS config used to connect to the sql database..
 	TLS `yaml:"tls_config"`
+
+	// The path to the database's CLI tool. This is only used for
+	// testing. Ex: /usr/bin/mysql
+	CLIPath string `yaml:"cli_path"`
```
daniel-nichter: Which CLI?

proto/s2s.go
```diff
@@ -133,6 +133,14 @@ func (js JobStatusByRuntime) Len() int           { return len(js) }
 func (js JobStatusByRuntime) Less(i, j int) bool { return js[i].Runtime > js[j].Runtime }
 func (js JobStatusByRuntime) Swap(i, j int)      { js[i], js[j] = js[j], js[i] }
 
+type JobLogs []JobLog
+
+func (jls JobLogs) Len() int { return len(jls) }
+func (jls JobLogs) Less(i, j int) bool {
+	return jls[i].RequestId+jls[i].JobId < jls[j].RequestId+jls[j].JobId
+}
+func (jls JobLogs) Swap(i, j int) { jls[i], jls[j] = jls[j], jls[i] }
```
daniel-nichter: I wonder if we're doing this "wrong". The Golang https://golang.org/pkg/sort/ pkg examples have type names like "FooByTs". So I wonder if it's more idiomatic to name these types as such (i.e. "...By...") to indicate that they're not just new types but specifically for sorting. What do you all think?

request-manager/jc/manager.go
```diff
@@ -0,0 +1,57 @@
+// Copyright 2017, Square, Inc.
+
+// Package jc provides and interface for managing Job Chains (JCs).
+package jc
```
daniel-nichter: When I first encountered "jc" way up there ^, I couldn't figure out what it meant. I had to search for this package. I guessed "jl" correctly. What about more explicit (but still short) package names? Future devs will thank us.

Side node: I still like like short var names inside the code--I think that's idiomatic for Go when used consistently. E.g. I've come to expect "rmc" and "jrc" to mean request manager client and job runner client.

request-manager/jl/manager.go
```diff
@@ -0,0 +1,197 @@
...
+		&jl.Stdout,
+		&jl.Stderr,
+	)
+	if err != nil {
+		return jl, err
+	}
+
+	// Update the finished job count on the request if the job completed successfully.
+	if jl.State == proto.STATE_COMPLETE {
+		q = "UPDATE requests SET finished_jobs = finished_jobs + 1 WHERE request_id = ?"
```
daniel-nichter: If job-log component is only responsible for job logs, should it deal with/affect requests directly?

request-manager/request/manager.go
```diff
@@ -1,106 +1,94 @@
...
-
 	*sync.Mutex
 }
 
-func NewManager(requestResolver *grapher.Grapher, dbAccessor DBAccessor, jrClient jr.Client) Manager {
+func NewManager(gr *grapher.Grapher, dbc db.Connector, jrClient jr.Client) Manager {
 	return &manager{
-		rr:    requestResolver,
-		db:    dbAccessor,
+		gr:    gr,
```
daniel-nichter: 👍 I like "gr" for grapher. We should use this consistently.

request-manager/test/db/db.go
```diff
@@ -0,0 +1,217 @@
...
+	defaultCmdArgs   []string
+	cliPath          string
+	//
+	dbs map[string]*sql.DB // db name => *sql.Db
+	*sync.Mutex
+}
+
+// A Manager manages the lifecycle of test databases. It is safe to use
+// with multiple databases concurrently.
+type Manager interface {
```
daniel-nichter: Good start on this. This is something we'll need to make db testing simple. I've been thinking that we might just want to factor out and abstract the whole concept, because we could use it for Etre and other various projects, too. But first, I'll look around and see if there's already a reusable solution. If not, I think we can make a good contribution to the Go+MySQL world with such a pkg, time permitting.

PR #38, opened 2018-01-30
Add job.Id struct and internally id jobs by Id.Id instead of Id.Name

daniel-nichter:
Also:
* Remove `grapher.Payload` and use `job.Job` instead, which is same interface
* Remove unused `graph` functions (dead code)
* Update copyright on modified files


job/job.go
```diff
@@ -63,23 +65,66 @@ type Job interface {
...
+// Id is defined per request by Spin Cycle. An example for each value:
+//
+//   Type:  net/down-ip
+//   Name:  down-sip-on-eth0
+//   Id:    9m4e2mr0
+//
+// Job types are defined in code and in the external job factory (EJF). They
+// represent what the job is and does. Job names are defined in request specs;
+// they are named instantiantions of job types. The name should be short but
+// meaningful the the context of the request. Job IDs are short, random strings
```
michaelfinch: meaningful in the*

job/job.go
```diff
@@ -63,23 +65,66 @@ type Job interface {
...
+// globally.
+type Id struct {
+	// User-defined low-level job type implemented in code and made by external
+	// job factory (EJF). Any value that the EJF is coded to make is valid. Job
+	// types are heavily reused by being given unique names in request specs.
+	Type string
 
-	// Type returns the type of the job.
-	Type() string
+	// User-defined job name in a request spec. Names must be unique per request.
```
michaelfinch: Probably more accurate to say that names must be unique per sequences, no? Two different sequences used within a request should both be able to have a job with the same name

PR #43, opened 2018-02-27
Add Retryable Sequences.

alyssa:


alyssa:
This is still a WIP, not ready for review yet.

proto/s2s.go
```diff
@@ -65,9 +72,10 @@ type RequestArg struct {
 // JobLog represents a log entry for a finished job.
 type JobLog struct {
 	// These three fields uniquely identify an entry in the job log.
-	RequestId string `json:"requestId"`
-	JobId     string `json:"jobId"`
-	Try       uint   `json:"try"` // try number N of 1 + Job.Retry
+	RequestId   string `json:"requestId"`
+	JobId       string `json:"jobId"`
+	Try         uint   `json:"try"`         // try number N of 1 + Job.Retry
```
daniel-nichter: Now that we have multiple "try", let's disambiguate this one: `JobTry`.

request-manager/grapher/spec.go
```diff
@@ -33,6 +33,7 @@ type SequenceSpec struct {
 	Name  string               `yaml:"name"`  // name of the sequence
 	Args  SequenceArgs         `yaml:"args"`  // arguments to the sequence
 	Nodes map[string]*NodeSpec `yaml:"nodes"` // list of nodes that are a part of the sequence
+	Retry uint                 `yaml:"retry"` // the number of times to retry a "sequence" that fails
```
daniel-nichter: Please excuse the minor non-code nit but why is "sequence" quoted? Is not a real sequence, or the same type of sequence previous fields refer to?

request-manager/resources/request_manager_schema.sql
```diff
@@ -24,18 +24,19 @@ CREATE TABLE IF NOT EXISTS `raw_requests` (
...
-  `started_at`  BIGINT UNSIGNED NOT NULL DEFAULT 0, -- Unix time (nanoseconds)
-  `finished_at` BIGINT UNSIGNED NOT NULL DEFAULT 0, -- Unix time (nanoseconds)
-  `error`       TEXT NULL DEFAULT NULL,
-  `exit`        TINYINT UNSIGNED NULL DEFAULT NULL,
-  `stdout`      LONGBLOB NULL DEFAULT NULL,
-  `stderr`      LONGBLOB NULL DEFAULT NULL,
+  `request_id`   BINARY(20) NOT NULL,
+  `job_id`       BINARY(4) NOT NULL,
+  `name`         VARBINARY(100) NOT NULL,
+  `try`          SMALLINT NOT NULL DEFAULT 0,
```
daniel-nichter: Disambiguate

request-manager/resources/request_manager_schema.sql
```diff
@@ -24,18 +24,19 @@ CREATE TABLE IF NOT EXISTS `raw_requests` (
...
+  `type`         VARBINARY(75) NOT NULL,
+  `state`        TINYINT UNSIGNED NOT NULL DEFAULT 0,
+  `started_at`   BIGINT UNSIGNED NOT NULL DEFAULT 0, -- Unix time (nanoseconds)
+  `finished_at`  BIGINT UNSIGNED NOT NULL DEFAULT 0, -- Unix time (nanoseconds)
+  `error`        TEXT NULL DEFAULT NULL,
+  `exit`         TINYINT UNSIGNED NULL DEFAULT NULL,
+  `stdout`       LONGBLOB NULL DEFAULT NULL,
+  `stderr`       LONGBLOB NULL DEFAULT NULL,
 
   PRIMARY KEY (`request_id`, `job_id`, `try`)
```
daniel-nichter: Does the PK need to change?

job-runner/chain/traverser.go
```diff
@@ -377,3 +386,36 @@ func (t *traverser) sendJL(pJob proto.Job, err error) {
 		jLogger.Errorf("problem sending job log (%#v) to the RM: %s", jl, err)
 	}
 }
+
+func (t *traverser) retrySequence(j proto.Job) {
```
daniel-nichter: Please avoid `j` because it's nearly impossible to follow or find in code. You'll notice `pJob` and `doneJ` (which should probably be `doneJob`). There's also `realJob` in  other code. These names are more illuminating and far easier to find/follow.

job-runner/chain/traverser.go
```diff
@@ -377,3 +386,36 @@ func (t *traverser) sendJL(pJob proto.Job, err error) {
...
+	sequenceStartJob := t.chain.SequenceStartJob(j)
+
+	// Create list of jobs in sequence that have been ran and we need to undo
```
daniel-nichter: What does it mean to "undo" a job? I have a pretty good idea, but what questions/possibilities might arise in the mind of future engineers? For example, "undo" could mean "rollback" somehow (as if jobs could undo their work in preparation for re-doing it).

job-runner/chain/traverser.go
```diff
@@ -377,3 +386,36 @@ func (t *traverser) sendJL(pJob proto.Job, err error) {
...
+	var jobsToUndo []string
+	jobsToUndo = append(jobsToUndo, j.Id)
+	currSeqJobId := sequenceStartJob.Id
+	for currSeqJobId != j.Id {
+		jobsToUndo = append(jobsToUndo, currSeqJobId)
+		nextJobs := t.chain.NextJobs(currSeqJobId)
+		if len(nextJobs) != 1 {
+			// error
```
daniel-nichter: :-)

job-runner/chain/traverser.go
```diff
@@ -377,3 +386,36 @@ func (t *traverser) sendJL(pJob proto.Job, err error) {
...
+		}
+		currSeqJobId = nextJobs[0].Id
```
daniel-nichter: Intention here (and this code block in general) is not clear to me. We're setting the current sequence job ID to the next job's ID... I guess I don't know what the purpose/intention of a "sequence job"?

job-runner/chain/traverser.go
```diff
@@ -377,3 +386,36 @@ func (t *traverser) sendJL(pJob proto.Job, err error) {
...
+		t.chain.SetJobState(jobId, proto.STATE_PENDING)
+	}
+
+	// Decrement count of jobs from total count of jobs done in chain
+	t.chain.n -= uint(len(jobsToUndo))
+
+	// Increment retry count for this sequence
+	t.chain.IncrementSequenceRetryCount(sequenceStartJob.Id)
+
+	t.sendJL(j, nil)
```
daniel-nichter: What's being logged here?

job-runner/chain/traverser.go
```diff
@@ -377,3 +386,36 @@ func (t *traverser) sendJL(pJob proto.Job, err error) {
...
+
+	// Enqueue first job in sequence
+	t.runJobChan <- sequenceStartJob
```
daniel-nichter: This could cause a panic. To see it, consider the implicit concurrency usage of this chan, and which chunk of code (a function) is responsible for the life and management, so to speak, of this chan.

proto/s2s.go
```diff
@@ -20,6 +20,13 @@ type Job struct {
 	Data      map[string]interface{} `json:"data"`      // job-specific data during Job.Run
 	Retry     uint                   `json:"retry"`     // retry N times if first run fails
 	RetryWait uint                   `json:"retryWait"` // wait time (milliseconds) between retries
+	Sequence  *Sequence              `json:"sequence"`
```
daniel-nichter: `// set if job begins or ends a sequence, else nil`

proto/s2s.go
```diff
@@ -20,6 +20,13 @@ type Job struct {
...
+}
+
+type Sequence struct {
```
daniel-nichter: `// Sequence represents ...`

Code docs more lax deeper in the code we go, but here at the very surface we need the best possible docs we can muster.

request-manager/grapher/graph.go
```diff
@@ -27,13 +27,16 @@ type Graph struct {
...
+	Datum           job.Job                // Data stored at this Node
+	Next            map[string]*Node       // out edges ( node id -> Node )
+	Prev            map[string]*Node       // in edges ( node id -> Node )
+	Name            string                 // the name of the node
+	Args            map[string]interface{} // the args the node was created with
+	Retry           uint                   // the number of times to retry a node
+	RetryWait       uint                   // the time, in seconds, to sleep between retries
+	SequenceStartId string
+	SequenceEndId   string
+	SequenceRetry   uint // only recorded on start node
```
daniel-nichter: I don't understand the code comment. What's "recorded"? And what does "on start node" mean--oh, the first node in the sequence?

job-runner/chain/chain.go
```diff
@@ -18,28 +18,41 @@ type chain struct {
...
+	sequenceRetryCount map[string]uint            // Number of retries attempted so far
 }
 
 // NewChain takes a JobChain proto (from the RM) and turns it into a Chain that
 // the JR can use.
 func NewChain(jc *proto.JobChain) *chain {
+	sequenceRetryCount := make(map[string]uint)
 	// Set the state of all jobs in the chain to "Pending".
 	for jobName, job := range jc.Jobs {
+		// Set the state of all jobs in the chain to "Pending".
```
michaelfinch: The same comment exists 2 lines above

job-runner/chain/chain.go
```diff
@@ -18,28 +18,41 @@ type chain struct {
...
 		job.State = proto.STATE_PENDING
+
+		// Initialize job.Data
 		job.Data = map[string]interface{}{}
+
+		// This is the start of a sequence that has can be retried. Let's initalize
+		// it count to 0 for now, then increment it when we do a retry
```
michaelfinch: that can be*
its count*

job-runner/chain/traverser.go
```diff
@@ -377,3 +386,36 @@ func (t *traverser) sendJL(pJob proto.Job, err error) {
...
 }
+
+func (t *traverser) retrySequence(j proto.Job) {
+	sequenceStartJob := t.chain.SequenceStartJob(j)
+
+	// Create list of jobs in sequence that have been ran and we need to undo
+	var jobsToUndo []string
+	jobsToUndo = append(jobsToUndo, j.Id)
+	currSeqJobId := sequenceStartJob.Id
+	for currSeqJobId != j.Id {
```
michaelfinch: Don't we want to include the current job (the one that failed and has prompted the sequence retry) in the jobsToUndo slice?

request-manager/grapher/spec.go
alyssa: It was copy/pasta from code comment above that says

```
// the number of times to retry a "job" that fails```

I'll remove the quotes

File: job-runner/chain/traverser.go (hunk 5 above)
Comment (alyssa) [reply job-runner/chain/traverser.go]: ok, assuming `pJob` is short of `protoJob` and not `pendingJob`...I did `s/j/pJob` for this. I also did `s/doneJ/doneJob`.

File: job-runner/chain/traverser.go (hunk 6 above)
Comment (alyssa) [reply job-runner/chain/traverser.go]: undo means to make the system think it has to run the job again by:

* making job pending again
* decrement those # of jobs from count in chain

I can rename this to `jobsToRetry`

File: job-runner/chain/traverser.go (hunk 16 above)
Comment (alyssa) [reply job-runner/chain/traverser.go]: yeah that's done two lines above. it's the first item in the array

Comment (alyssa) [reply job-runner/chain/traverser.go]: let me add a code comment to make this more clear

File: request-manager/grapher/grapher.go [hunk 17]
```diff
@@ -307,9 +307,27 @@ func (o *Grapher) buildComponent(name string, nodeDefs map[string]*NodeSpec, nod
...
+		// and once when putting the entire graph together. We only want to mark
+		// the Sequence{End,Start}Id when constructing the component.
+		if vertex.SequenceStartId == "" && vertex.SequenceEndId == "" {
+			vertex.SequenceStartId = sequenceStartId
+			vertex.SequenceEndId = sequenceEndId
+		}
+	}
+
 	// Assert g is a well formed graph
 	if !g.IsValidGraph() {
```
BrianIp: Request: you can update IsValidGraph() to check that SequenceStartId and SequenceEndId match what we expect them to be. Not a hard blocker, but would be useful if/when we need to update graphing code in the future.

BrianIp: Request: you can update IsValidGraph() to check that SequenceStartId and SequenceEndId match what we expect them to be. Not a hard blocker, but would be useful if/when we need to update graphing code in the future.

request-manager/resources/request_manager_schema.sql
alyssa: oh good catch. yes, adding `sequence_try`

job-runner/chain/traverser.go
alyssa: actually i'll rename `j` to `failedJob`

job-runner/chain/traverser.go
alyssa: it's a job in the sequence. i've reworked the loop to make it more clear.

proto/s2s.go
alyssa: This is actually set for all jobs. I'll update comment to say that.

proto/s2s.go
alyssa: Added a comment. TBH I was just writing code and forgot to come back and update comments.

request-manager/grapher/graph.go
alyssa: yeah the first node in the sequence. i'll update

job-runner/chain/chain.go
alyssa: removed the comment 2 lines above

job-runner/chain/chain.go
alyssa: updated

request-manager/resources/request_manager_schema.sql
alyssa: This involves renaming a column which I think the process is:

1. create a new column called job_try and change code to also set job_try
2. once deployed internally, do a backfill of try -> job_try
3. Delete `try` column

Given that the work for this takes some time, I'd rather do it in a separate PR so that it's clear where the change starts and ends.

proto/s2s.go
alyssa: I think this rename is best to do in another PR. See my comment here: https://github.com/square/spincycle/pull/43#discussion_r172871963

job-runner/chain/traverser.go
alyssa: The failed job is being logged here, which also includes the sequence try it was on. I've renamed `j` to `failedJob` which hopefully clears that up.

request-manager/grapher/grapher.go
alyssa: IsValidGraph() seems to check for the formation of the graph (the edges, vertices) but not the data in the actual vertices, which is what I changed. So, I don't think it's the correct place to check? I'll be adding a test to make sure the correct data is set

job-runner/chain/chain.go
```diff
@@ -18,28 +18,40 @@ type chain struct {
...
-	// Set the state of all jobs in the chain to "Pending".
+	sequenceRetryCount := make(map[string]uint)
 	for jobName, job := range jc.Jobs {
+		// Set the state of all jobs in the chain to "Pending".
 		job.State = proto.STATE_PENDING
+
+		// Initialize job.Data
 		job.Data = map[string]interface{}{}
+
+		if job.Sequence.Retry > 0 && jobName == job.Sequence.StartJobId {
```
alyssa: In order to support nested sequence retries, we need to change the way we store info about the Sequence retries. 

Since every job is part of a sequence, we can change to encode parent sequence start id and on the start node of ever sequence, encode the actual retry.

This is different from above because the start node will no longer necessarily have the same jobeName job.Sequence.StartJobId

request-manager/resources/request_manager_schema.sql
daniel-nichter: Sure, we can do another PR, but at this stage of development we don't really need to do a full migration and backfill. We haven't released any stable versions yet.

job-runner/chain/traverser.go
```diff
@@ -377,3 +400,47 @@ func (t *traverser) sendJL(pJob proto.Job, err error) {
...
+	for currJob.Id != failedJob.Id {
+		// Mark failed job as pending and update chain job count
+		t.chain.SetJobState(currJob.Id, proto.STATE_PENDING)
+		t.chain.n -= 1
+
+		// Update currJob to next job in sequence
+		nextJobs := t.chain.NextJobs(currJob.Id)
+		if len(nextJobs) != 1 {
+			err := fmt.Errorf("traversing sequence and expected next jobs list to be length of 1 but it is lenght %d", len(nextJobs))
+			return err
```
daniel-nichter: `return fmt.Errorf(...`

job-runner/chain/traverser.go
```diff
@@ -377,3 +400,47 @@ func (t *traverser) sendJL(pJob proto.Job, err error) {
...
+	currJob := sequenceStartJob
+	for currJob.Id != failedJob.Id {
+		// Mark failed job as pending and update chain job count
+		t.chain.SetJobState(currJob.Id, proto.STATE_PENDING)
+		t.chain.n -= 1
+
+		// Update currJob to next job in sequence
+		nextJobs := t.chain.NextJobs(currJob.Id)
+		if len(nextJobs) != 1 {
+			err := fmt.Errorf("traversing sequence and expected next jobs list to be length of 1 but it is lenght %d", len(nextJobs))
```
daniel-nichter: Is this really possible? This is the only possible error return in this func, but we already have a known-good job chain that's already been traversed so... why are we checking for this?  Because if we not check for errors, then it simplifies the caller a little, too (i.e. my previous comment re logging seq retry before calling this func).

job-runner/chain/traverser.go
```diff
@@ -201,7 +201,20 @@ JOB_REAPER:
 
 		// If the job did not complete successfully, then ignore subsequent jobs.
 		// For example, if job B of A -> B -> C  fails, then C is not ran.
-		if doneJ.State != proto.STATE_COMPLETE {
+		if doneJob.State != proto.STATE_COMPLETE {
+			// TODO(alyssa): Add supported for retrying nested sequences. This only
+			// retries current sequence we are in.
+			if t.chain.CanRetrySequence(doneJob) {
+				jLogger.Info("job did not complete successfully. retrying sequence.")
+				t.sendJL(doneJob, nil)
```
michaelfinch: I don't think this is needed. The job-runner will always create a job log when it runs a job. The only reason we need to create job logs in the traverser is when we fail to create a job runner or some higher-level thing like that

job-runner/chain/traverser.go
```diff
@@ -377,3 +395,47 @@ func (t *traverser) sendJL(pJob proto.Job, err error) {
 		jLogger.Errorf("problem sending job log (%#v) to the RM: %s", jl, err)
 	}
 }
+
+// retrySequence retries a sequence by doing the following
+//   1. Mark the failed job and all jobs before it in the sequence to PENDING
+//   state, so that they are ran again on sequence retry. This is OK to do
+//   since all jobs are idempotent.
```
michaelfinch: Fwiw, there is currently no mention of jobs being idempotent in the design doc. I know we generally try to make our jobs that way, but I'm not sure we've officially agreed that all jobs created for spincycle should be idempotent

job-runner/chain/traverser.go
alyssa: ah yes the runner does create a job log. ok i'll remove this and update the runner to include the sequence related fields. good catch - thanks.

job-runner/chain/traverser.go
alyssa: Hm, ok. Thinking about this a bit and there are a few options:

1) when we configure a sequence retry, ensure that all jobs it may affect are idempotent. However, someone may forget to check or we may later add a job to the retryable sequence that's not idempotent. Given that nothing can be ensured, I don't really like this best effort / check as we go approach. We must be doing this approach for job retries, though.
2) Make it a requirement that all jobs are idempotent. This would require doing an audit before merging in this change and updating the design docs. I'm more of a fan of this one.

WDYT?

job-runner/chain/traverser.go
```diff
@@ -201,7 +201,25 @@ JOB_REAPER:
...
+		if doneJob.State != proto.STATE_COMPLETE {
+			// Check that the job has failed or timed out and we are able to retry
+			// the sequence. Note that exclude STATE_UNKNOWN and STATE_STOPPED from
+			// this list.
+			// TODO(alyssa): This only retries current sequence we are
+			// in. Add supported for retrying nested sequences.
+			if (doneJob.State == proto.STATE_FAIL ||
+				doneJob.State == proto.STATE_TIMEOUT) &&
+				t.chain.CanRetrySequence(doneJob) {
+				jLogger.Info("job did not complete successfully. Retrying sequence.")
```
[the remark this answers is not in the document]
alyssa: this comment got lost but i just found it. good call. moved the log line.

job-runner/chain/traverser.go
alyssa: we chatted offline. it's possible for the sequence retry to include a nested sequence. removed this length check.

job-runner/chain/chain.go
```diff
@@ -18,28 +18,39 @@ type chain struct {
 	jcMux    *sync.RWMutex
 	jobChain *proto.JobChain
 
-	runningMux *sync.RWMutex
-	running    map[string]proto.JobStatus // keyed on job id
-	n          uint                       // N jobs ran
+	runningMux         *sync.RWMutex
+	running            map[string]proto.JobStatus // keyed on job id
+	n                  uint                       // N jobs ran
+	sequenceRetryCount map[string]uint            // Number of retries attempted so far
```
alyssa: I do feel like this name is too similar to `job.RetryCount`

`job.RetryCount` - configured number of sequence retries
`sequenceRetryCount` - number of retries attempted so far

so one is static and the other might vary. given this, i don't think i should call them both `RetryCount`

`sequenceRetryNumber` for the running count?

request-manager/resources/request_manager_schema.sql
```diff
@@ -24,18 +24,20 @@ CREATE TABLE IF NOT EXISTS `raw_requests` (
...
-  `finished_at` BIGINT UNSIGNED NOT NULL DEFAULT 0, -- Unix time (nanoseconds)
-  `error`       TEXT NULL DEFAULT NULL,
-  `exit`        TINYINT UNSIGNED NULL DEFAULT NULL,
-  `stdout`      LONGBLOB NULL DEFAULT NULL,
-  `stderr`      LONGBLOB NULL DEFAULT NULL,
+  `request_id`   BINARY(20) NOT NULL,
+  `job_id`       BINARY(4) NOT NULL,
+  `name`         VARBINARY(100) NOT NULL,
+  `try`          SMALLINT NOT NULL DEFAULT 0,
+  `sequence_id`  BINARY(4) NOT NULL, -- TODO(alyssa) figure out default value
```
alyssa: it's possible this may be empty, for example the first job in the graph does not have a parent sequence id. idk what the default should be. can i just put `DEFAULT 0`? @michaelfinch

alyssa: I thought about this some more and i think this exposes a design flaw in my solution. all jobs are part of a sequence by definition, so there should be no case in which a job does not have a sequence id. to solve this, i will have `sequence_id` and `parent_sequence_id`. this just becomes a little tricky when i record the job in the job log. if a job is part of a sub sequence in its parent sequence, but we retry the parent sequence, we want to log that this job is being retried as part of the parent sequence. will look at code and see how to make it happen.

job-runner/chain/chain.go
daniel-nichter: There's no `job. RetryCount` afaik?  `sequenceRetryCount` makes sense to me, more than `sequenceRetryNumber`.

job-runner/chain/traverser.go
daniel-nichter: I think this is a documentation issue for elsewhere, isn't a matter for this private function. For user-facing docs, we should say that jobs _should_ support being retried (being truly idempotent is a strong requirement; is possible to support retry with slightly different results on retry, which means it's not technical idempotent). But really it's up to user because it's not something Spin Cycle itself can ever check or enforce. We just run the jobs... up to user to write code to handle retries if they use retries.

job-runner/chain/traverser.go
```diff
@@ -377,3 +395,85 @@ func (t *traverser) sendJL(pJob proto.Job, err error) {
...
+	// Increment retry count for this sequence
+	t.chain.IncrementSequenceRetryCount(failedJob)
+
+	// Enqueue first job in sequence
+	t.runJobChan <- sequenceStartJob
+
+	return nil
+}
+
+// jobIdsToRetry does a breadth first search to find all jobs whose state is not PENDING, which means the state is either COMPLETE or FAILED. This is a list of jobs that need to be retried.
```
daniel-nichter: Nit: let's break this line so we don't have to scroll horizontally. We usually line break around 78-80 chars.

job-runner/chain/traverser.go
```diff
@@ -377,3 +395,85 @@ func (t *traverser) sendJL(pJob proto.Job, err error) {
...
+func (t *traverser) jobIdsToRetry(sequenceStartJobId string) []string {
+	// visited is a list of job ids visited
+	var visited []string
+	// frontier is a list of job ids to visit
+	var frontier []string
```
daniel-nichter: s/frontier/toVisit/  or /pending/? Unless "frontier" is normal terminology for this type of code.

job-runner/chain/traverser.go
```diff
@@ -377,3 +395,85 @@ func (t *traverser) sendJL(pJob proto.Job, err error) {
...
+
+	for _, pJob := range t.chain.NextJobs(sequenceStartJobId) {
+		frontier = append(frontier, pJob.Id)
+	}
+	visited = append(visited, sequenceStartJobId)
+
+	// Search while the frontier set is non-empty
+	for len(frontier) > 0 {
+
+		// For every node in the frontier
```
daniel-nichter: Signal to noise ratio wrt comments

job-runner/chain/traverser.go
```diff
@@ -377,3 +395,85 @@ func (t *traverser) sendJL(pJob proto.Job, err error) {
...
+	t.chain.IncrementSequenceRetryCount(failedJob)
+
+	// Enqueue first job in sequence
+	t.runJobChan <- sequenceStartJob
+
+	return nil
+}
+
+// jobIdsToRetry does a breadth first search to find all jobs whose state is not PENDING, which means the state is either COMPLETE or FAILED. This is a list of jobs that need to be retried.
+func (t *traverser) jobIdsToRetry(sequenceStartJobId string) []string {
```
daniel-nichter: Little difficult to follow this function, but if I'm reading it correctly and we have:
`A > [B > C] > D > E`
where `[ ]` denotes a sequence, and `C` fails, then `sequenceStartJobId=B` and we walk the graph all the way to `E`?   If true, we're doing too much work because `D` and `E` don't need to be reset.

In the previous incarnation of this algo (the simpler one in `retrySequence()` we stopped walking at `C`.

daniel-nichter: Granted, it's late (11:30pm) my time, and I'm new to this code, but I cannot follow this function even though I know its intention. We should do a code walkthrough through with @BrianIp since he did similar code in grapher. For example, what boggles me is that we loop on `frontier` but in that loop we both add to _and_ delete from `frontier`.

job-runner/runner/runner.go
```diff
@@ -39,33 +39,37 @@ type runner struct {
...
+	stopChan    chan struct{}
 	*sync.Mutex
 	logger    *log.Entry
 	startTime time.Time
 }
 
 // NewRunner takes a proto.Job struct and its corresponding job.Job interface, and
 // returns a Runner.
-func NewRunner(pJob proto.Job, realJob job.Job, reqId string, rmc rm.Client) Runner {
+func NewRunner(pJob proto.Job, realJob job.Job, reqId string, sequenceTry uint, rmc rm.Client) Runner {
```
daniel-nichter: I think this begins to leak abstraction. I know we need seq id/try for the JLE, but imho job runner should not be aware of sequences, which are a higher abstraction. So in cases like this, "most problems can be solved by another level of abstraction". What we want is a thing to do logging, and we can give that thing all the info it needs (e.g. seq id/try), and then pass that thing into the job runner. So the job runner gets logging capabilities without any awareness of things it should be aware of.

request-manager/resources/request_manager_schema.sql
daniel-nichter: Does `sequence_id` tell us anything meaningful, i.e. to a human? Job IDs are mostly a necessity for establishing a short primary key to identify rows in this table, but to a human, what about a sequence do we want to know wrt a job? How would I ever look up or know what `sequence_id=x9r4` is?

request-manager/resources/request_manager_schema.sql
```diff
@@ -24,18 +24,20 @@ CREATE TABLE IF NOT EXISTS `raw_requests` (
...
+  `state`        TINYINT UNSIGNED NOT NULL DEFAULT 0,
+  `started_at`   BIGINT UNSIGNED NOT NULL DEFAULT 0, -- Unix time (nanoseconds)
+  `finished_at`  BIGINT UNSIGNED NOT NULL DEFAULT 0, -- Unix time (nanoseconds)
+  `error`        TEXT NULL DEFAULT NULL,
+  `exit`         TINYINT UNSIGNED NULL DEFAULT NULL,
+  `stdout`       LONGBLOB NULL DEFAULT NULL,
+  `stderr`       LONGBLOB NULL DEFAULT NULL,
 
-  PRIMARY KEY (`request_id`, `job_id`, `try`)
+  PRIMARY KEY (`request_id`, `job_id`, `try`, `sequence_id`, `sequence_try`)
```
daniel-nichter: Related to comment above, is `sequence_id` necessary to unique identify a row? A PK must be only necessary and sufficient cols to unique ID each row.

job-runner/chain/chain.go
alyssa: There is `job.RetryCount`. `RetryCount` field is only set on the job that is the start of the sequence. https://github.com/square/spincycle/blob/c93f793dedec8d682acb2343637bfdba3a20aedd/proto/s2s.go#L24

request-manager/resources/request_manager_schema.sql
alyssa: a job can be part of many sequences. it may be retried in the immediate sequence that it is part of or it may be retried as part of one of its parent's sequences. so, a job sequence retry count meaningless unless we know what sequence initiated the retry. This will help us differentiate a single job that was part of distinct sequence retry runs.

job-runner/chain/traverser.go
alyssa: frontier is normal terminology for breadth first search and the same verbiage used in a different part in this repo. 

https://github.com/square/spincycle/blob/1e3a3845f4c6fdc0ca9974fac2aad276673f32d8/request-manager/grapher/graph.go#L241

request-manager/resources/request_manager_schema.sql
alyssa: answered in comment above

job-runner/chain/traverser.go
alyssa: I've renamed it to `toVisit`. although frontier is normal terminology it did confuse you and it confused me at first until i googled it. so we might as well go for the more intuitive variable name since it has the same implication.

job-runner/chain/traverser.go
```diff
@@ -199,9 +199,21 @@ JOB_REAPER:
...
 		}
 
-		// If the job did not complete successfully, then ignore subsequent jobs.
-		// For example, if job B of A -> B -> C  fails, then C is not ran.
-		if doneJ.State != proto.STATE_COMPLETE {
+		// t.runJobs() sets every job that isn't COMPLETE to be FAIL.
+		if doneJob.State != proto.STATE_COMPLETE {
+			// If the job is part of a retryable sequence, retry that sequence.
+			if t.chain.CanRetrySequence(doneJob) {
+				err := t.retrySequence(doneJob)
```
daniel-nichter: Since `retrySequence` cannot fail (return error), let's make it not return anything. The error handling here is superfluous. Without it, the block is much cleaner imho:
```go
jLogger.Info("job did not complete successfully. retrying sequence.")
t.retrySequence(doneJob) // re-enqueue first job of failed sequence
continue JOB_REAPER
```

job-runner/chain/traverser.go
```diff
@@ -360,15 +373,21 @@ func (t *traverser) runJobs() {
...
-		FinishedAt: 0,
-		State:      pJob.State,
-		Exit:       1,
-		Error:      err.Error(),
+		RequestId:   t.chain.RequestId(),
+		JobId:       pJob.Id,
+		Name:        pJob.Name,
+		Type:        pJob.Type,
+		Try:         1 + t.chain.SequenceRetryCount(pJob),
+		SequenceTry: 1 + t.chain.SequenceRetryCount(pJob),
```
daniel-nichter: How can both `Try` and `SequenceTry` be the same value? Also, there was no `Try` previously (in diff).

daniel-nichter: And `Error:      err.Error(),` was removed?

job-runner/chain/traverser.go
```diff
@@ -377,3 +396,99 @@ func (t *traverser) sendJL(pJob proto.Job, err error) {
...
+
+			continue PROCESS_CURRENT_JOB
+		}
+
+		continue PROCESS_TO_VISIT_LIST
+	}
+
+	var completedJobs []proto.Job
+	for _, j := range visited {
+		completedJobs = append(completedJobs, j)
```
daniel-nichter: Optimization: whenever we know the size of an array, we should allocate only that size:
```go
completedJobs := make([]proto.Job, 0, len(visited))
```
This prevents Go from doing dynamic allocations in bigger chunks that often go unused. Obviously not a huge pefr win, but it's easy enough to implement, so might as well.

job-runner/runner/runner.go
```diff
@@ -106,18 +110,20 @@ TRY_LOOP:
...
-			State:      jobRet.State,
-			Exit:       jobRet.Exit,
-			Error:      errMsg,
-			Stdout:     jobRet.Stdout,
-			Stderr:     jobRet.Stderr,
+			RequestId:   r.reqId,
+			JobId:       r.jobId,
+			Name:        r.jobName,
+			Type:        r.jobType,
+			Try:         tryNo + r.sequenceRetry,
```
daniel-nichter: This strikes me as wrong? Job try should be monotonically increasing. With this, we'll get:
1
2
4
For <job,seq> retry <1,1>, <2,1>, <3,1>

proto/s2s.go
```diff
@@ -65,9 +67,11 @@ type RequestArg struct {
 // JobLog represents a log entry for a finished job.
 type JobLog struct {
 	// These three fields uniquely identify an entry in the job log.
-	RequestId string `json:"requestId"`
-	JobId     string `json:"jobId"`
-	Try       uint   `json:"try"` // try number N of 1 + Job.Retry
+	RequestId   string `json:"requestId"`
+	JobId       string `json:"jobId"`
+	Try         uint   `json:"try"`         // try number N of 1 + Job.Retry + SequenceTry
```
daniel-nichter: Same issue here wrt job try being monotonic

request-manager/grapher/grapher_test.go
```diff
@@ -89,6 +89,9 @@ func TestNodeArgs(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	// arp debugging
+	g.PrintDot()
```
daniel-nichter: ⭐️ Need to remove this

job-runner/runner/runner.go
daniel-nichter: Oh actually, it's more like  <1,0>, <2,0>, <3,1> so 1, 2, 4 (mixing tryNo with retry). In any case: seq try/retry counts shouldn't factor into the job try count, which should always be monotonic starting from 1.

job-runner/chain/traverser.go
alyssa: good point. i must have been returning an error before and removed it. thank you

job-runner/chain/traverser.go
alyssa: cool. good idea. updated it to be `make([]proto.Job, len(visited))` since we'll def fill the capacity.

request-manager/grapher/grapher_test.go
alyssa: removed

job-runner/runner/runner.go
alyssa: this is inside a loop of retrying the job. so the try number math was being done outside of the job log. i think the fix is to remove setting try no here and just letting job log handle it

alyssa: fixed this. we chatted offline but for record: job log isn't the source of truth. so to fix, pass in previous job try count and have the traverser be the source of truth.

proto/s2s.go
daniel-nichter: `s/ + SequenceTry//`

job-runner/runner/runner.go
```diff
@@ -145,6 +159,8 @@ TRY_LOOP:
 			break TRY_LOOP
 		}
 
+		tryNo++
```
daniel-nichter: Put this after the `// Wait between retries` block because in the rare case that the job is stopped while waiting, this would cause the wrong job_try to be reported. For example, let's say maxTries = 2 and try 1 fails, then we tryNo++ here so now job_try=2, then we wait, but the job is stopped during that wait, we'll return Tries=2 when only 1 try happened.

proto/s2s.go
alyssa: I replaced it with `try number that is monotonically increasing` because `1 + Job.Retry` is no longer true (can be greater than 1+ Job.Retry if the job maxes out retries then it tried again as part of a  sequence retry).

job-runner/runner/runner.go
alyssa: oof good catch. thank you.

PR #47, opened 2018-07-03
Lkim/static args

laurenkim1:


request-manager/request/manager.go
```diff
@@ -420,6 +421,17 @@ func (m *manager) Specs() []proto.RequestSpec {
 				Name:     arg.Name,
 				Desc:     arg.Desc,
 				Required: false,
+				Static:   false,
+				Default:  arg.Default,
+			}
+			s.Args = append(s.Args, a)
+		}
+		for _, arg := range req[name].Args.Static {
```
michaelfinch: I think it would better to actually for the request manager to not expose static args at all. Since clients can't update them anyway, they don't even need to know about them

dev/rm-config.yaml
```diff
@@ -9,4 +9,4 @@ db:
   dsn: "root:@tcp(127.0.0.1:3306)/spincycle_dev"
 jr_client:
   server_url: "http://127.0.0.1:9999"
-spec_file_dir: "../specs/"
+spec_file_dir: "specs/"
```
michaelfinch: Could you add a 5th assumption to the top of the README (but put it right after the jobs symlink assumption) that says something like "The specs symlink points to the specs you want (default: dev/specs)"

PR #49, opened 2018-07-11
Lkim/job status

laurenkim1:


request-manager/request/manager.go
```diff
@@ -542,6 +542,24 @@ func (m *manager) getWithJc(requestId string) (proto.Request, error) {
 	}
 
 	req.JobChain = &jc
+
+	var params proto.CreateRequestParams
+	var rawParams []byte // raw job chains are stored as blobs in the db.
+	q = "SELECT request FROM raw_requests WHERE request_id = ?"
```
michaelfinch: How could we get this info without making an extra call to the database?

laurenkim1: I thought the next line was a call to the database?

spinc/cmd/status.go
```diff
@@ -40,7 +40,21 @@ func (c *Status) Run() error {
 		return nil
 	}
 
+	running := []string{}
```
michaelfinch: make this a map of [jobName] => status instead?

spinc/cmd/status.go
```diff
@@ -40,7 +40,21 @@ func (c *Status) Run() error {
...
+	statuses := []string{}
```
michaelfinch: instead of using two slices, you could make a map of [jobName] => status

PR #50, opened 2018-07-11
Lkim/sequence retry

laurenkim1:


michaelfinch:
Does this PR address the ticket to only exposes things marked as `request` to end users, or the ticket to move sequence retries up to the caller, or both?

request-manager/grapher/spec.go
```diff
@@ -30,10 +30,11 @@ type NodeArg struct {
...
 // define each sequence
 type SequenceSpec struct {
-	Name  string               `yaml:"name"`  // name of the sequence
-	Args  SequenceArgs         `yaml:"args"`  // arguments to the sequence
-	Nodes map[string]*NodeSpec `yaml:"nodes"` // list of nodes that are a part of the sequence
-	Retry uint                 `yaml:"retry"` // the number of times to retry the sequence if it fails
+	Name    string               `yaml:"name"`    // name of the sequence
+	Args    SequenceArgs         `yaml:"args"`    // arguments to the sequence
+	Nodes   map[string]*NodeSpec `yaml:"nodes"`   // list of nodes that are a part of the sequence
+	Retry   uint                 `yaml:"retry"`   // the number of times to retry the sequence if it fails
```
michaelfinch: Since we don't want to allow retries to be set on sequence specs directly, remove the `yaml:"retry"` here.

PR #53, opened 2018-07-13
Lkim/cleanup

laurenkim1:
fix ps formatting


spinc/config/config.go
```diff
@@ -98,7 +98,7 @@ func Help(full bool, rmc rm.Client) {
 			"  --help     Print help\n"+
 			"  --ping     Ping addr\n"+
 			"  --timeout  API timeout, milliseconds (default: %d)\n"+
-			"  --version  Print version\n\n"+
+			"  --version  Print version\n"+
 			"  -v         verbose ps or status\n\n"+
```
michaelfinch: To make this consistent, can we have it be ` --verbose   Verbose ps...`

PR #54, opened 2018-07-31
Lkim/conditional paths

laurenkim1:


request-manager/grapher/grapher.go
```diff
@@ -181,13 +181,47 @@ func (o *Grapher) buildComponent(name string, nodeDefs map[string]*NodeSpec, nod
 
 				// Build next graph component and assert that it's valid
 				var g *Graph
-				if !n.isSequence() {
-					// Node is a job, create a graph that contains only the node
-					g, err = o.buildSingleVertexGraph(n, nodeArgsCopy)
+				if n.isConditional() {
+					// Node is a conditional, check the value of the "if" jobArg
```
BrianIp: This block should go into its own "buildConditional" function

laurenkim1: ok it's done

PR #57, opened 2018-08-30
suspend job chains on JR shutdown

felixplajer:
JR catches shutdown signals and notifies api + traversers:
- api refuses to start new job chains
- traverser suspends the running job chain and sends SJC via RM Client

(RM Client doesn't actually send the SJC to the RM - breaking the RM side of things up into a different PR)


job-runner/api/api.go
```diff
@@ -30,6 +30,9 @@ var (
 	ErrDuplicateTraverser = errors.New("traverser already exists")
 	ErrTraverserNotFound  = errors.New("traverser not found")
 	ErrInvalidTraverser   = errors.New("traverser found, but type is invalid")
+
+	// Error when Job Runner is shutting down and not starting new job chains
+	ErrNoNewJobChains = errors.New("Job Runner is shutting down - no new job chains are being started")
```
daniel-nichter: I think `ErrShuttingDown` is more clear, since that's what's happening. Not accepting new job chains is a side-effecting of shutting down.

job-runner/chain/traverser.go
```diff
@@ -503,3 +547,68 @@ PROCESS_TO_VISIT_LIST:
...
+		}
+		if err := t.rmc.SuspendRequest(requestId, sjc); err != nil {
+			t.logger.Errorf("problem sending suspended job chain to RM: %s", err)
+		} else {
+			t.chainRepo.Remove(t.chain.RequestId())
+		}
+	}
+}
+
+func (t *traverser) suspended() bool {
```
daniel-nichter: For consistency, I'd rename to `shuttingDown` like the func in api since it's the same code/test on `shutdownChan`. Plus, the JR Itself isn't really suspended, only job chains are suspended.

job-runner/chain/traverser.go
```diff
@@ -146,18 +165,31 @@ func (t *traverser) Run() error {
...
 		}
+
+		// Job chain is done running - doneChan is watched by suspend().
+		close(t.doneChan)
+	}()
+
+	// Watch for shutdown signal and suspend job chain when received.
+	go func() {
+		<-t.shutdownChan
+		t.suspend()
```
daniel-nichter: Same as a comment below: "shutdown" makes more sense to me because we're not suspending the traverser (we suspend job chains).  I.e. trying to keep `suspend*` a term related just to job chains, not to other things.

job-runner/server/server.go
```diff
@@ -17,91 +20,107 @@ import (
...
-	s.api = api
 	return nil
 }
 
 func (s *Server) API() *api.API {
 	return s.api
 }
 
+// Catch TERM and INT signals to gracefully shut down the Job Runner
+func (s *Server) StartSignalHandler() {
```
daniel-nichter: https://hackmongo.com/page/golang-antipatterns/#hidden-goroutine  :-)

job-runner/chain/chain.go
```diff
@@ -145,6 +145,8 @@ LOOP:
 			if c.CanRetrySequence(job) {
 				return false, false
 			}
+		case proto.STATE_STOPPED:
+			// Like STATE_FAIL, but ignore sequence retries.
```
daniel-nichter: We can `return true, false` here, right? If any job was stopped, then we know the chain is done (`true`) but impossible for it to be completed (`false`) because of the stopped job.

job-runner/chain/traverser.go
daniel-nichter: Hm, one second reading maybe this is ok. Still thinking about it. It helps to think of traverser + chain as one thing.

job-runner/chain/traverser.go
```diff
@@ -146,18 +165,31 @@ func (t *traverser) Run() error {
...
+			// because next we'll finalize the request in the RM. Although
+			// short-lived, we set it in case there's problems finalizing with RM.
+			t.chain.SetState(finalState)
+			t.chainRepo.Set(t.chain)
+
+			// Mark the request as finished in the Request Manager.
+			if err := t.rmc.FinishRequest(t.chain.RequestId(), finalState); err != nil {
+				t.logger.Errorf("problem reporting status of the finished chain: %s", err)
+			} else {
+				t.chainRepo.Remove(t.chain.RequestId())
```
daniel-nichter: For this blocks, let's add a `@todo: retry FinishRequest`.  I think we should always remove completed chains from repo, but in case RM is down or being redeployed and it temporarily fails, we should retry a few times. Right now, if it fails and we just log and leave chain in repo, I think that leaves phantom chains that are neither complete according to RM nor gone from chain repo.

job-runner/chain/chain.go
felixplajer: Yeah, `IsDone` could use some refining. I had some problems changing it too much because the traverser currently uses it to break out of the job reaper loop (<- might have to change how that works).

Question: If there's both a stopped job and a running job, is the chain "done"?

PR #59, opened 2018-09-26
JR: suspend + resume job chains

felixplajer:
- refactor the traverser, adding a new JobReaper to handle all done jobs
- suspend all running job chains when the JR is shutting down and send SJCs to the RM (well, the RM client doesn't actually send the SJC yet, since the RM code to handle it isn't there yet)
- receive an SJC from the RM and resume running that job chain


job-runner/chain/chain.go
```diff
@@ -97,103 +135,197 @@ func (c *chain) NextJobs(jobId string) proto.Jobs {
...
-	for _, job := range c.previousJobs(jobId) {
-		if job.State != proto.STATE_COMPLETE {
-			return false
+	return c.jobIsReady(jobId)
+}
+
+// RunnableJobs returns a list of all jobs that are ready to run. A job is
+// ready to run if all of its previous jobs are complete and it is Pending
+// or Stopped with some retries still remaining.
+func (c *chain) RunnableJobs() proto.Jobs {
```
daniel-nichter: Why does this not return `[]proto.Job`?

job-runner/chain/chain.go
```diff
@@ -97,103 +135,197 @@ func (c *chain) NextJobs(jobId string) proto.Jobs {
...
+func (c *chain) IsDoneRunning() (done bool, complete bool) {
 	c.jcMux.RLock()
 	defer c.jcMux.RUnlock()
 
 	done = true
 	complete = true
 	pendingJobs := proto.Jobs{}
 
+	// If any jobs are still running, the chain isn't done or complete.
+	for _, job := range c.jobChain.Jobs {
```
daniel-nichter: This block is not necessary, redundant. The same check and return is done in the next block:
```go
        case proto.STATE_RUNNING:
            // We should never reach this case because we previously check for running jobs.
            return false, false
```

job-runner/chain/chain.go
```diff
@@ -97,103 +135,197 @@ func (c *chain) NextJobs(jobId string) proto.Jobs {
...
-			}
-		}
-
-		// If all of the previous jobs of a pending job are complete, the
-		// chain can't be complete because the pending job can still run.
-		if allPrevComplete == true {
-			return false, complete
+		if c.jobIsReady(job.Id) {
+			// If a job can be run, the chain can't be done.
+			done = false
```
daniel-nichter: Might as well early return here if we're certain we know the final results. In such cases (early return), it prevents accidental code after the block, etc. Who knows what another dev might put after this block.

job-runner/chain/chain.go
```diff
@@ -97,103 +135,197 @@ func (c *chain) NextJobs(jobId string) proto.Jobs {
...
+			break
```
daniel-nichter: Might as well early return here if we're certain we know the final results. In such cases (early return), it prevents accidental code after the block, etc. Who knows what another dev might put after this block.

job-runner/chain/chain.go
```diff
@@ -15,19 +15,29 @@ import (
 
 // chain represents a job chain and some meta information about it.
 type chain struct {
+	// A note on the read/write mutexes in chain:
+	//   A lot of the methods in chain call each other - be careful not to make
+	//   nested RLock calls within one goroutine, or a Lock call from another
+	//   goroutine can cause a deadlock (and it will be very annoying to debug).
 	jcMux    *sync.RWMutex
```
daniel-nichter: All the mutexes are confusing. Can a single mutex guard any and all changes to the chain? Or would that serialize access too severely?

job-runner/chain/chain.go
```diff
@@ -340,16 +485,70 @@ func (c *chain) UnmarshalJSON(bytes []byte) error {
...
+// ready to run if it is Pending or Stopped with some retry attempts remaining,
+// and all of its previous jobs are complete. If any previous jobs are not
+// complete, the job is not ready to run.
+// Doesn't lock jcMux, so suitable to call within other chain methods lock jcMux.
+func (c *chain) jobIsReady(jobId string) bool {
+	// Only Pending or Stopped jobs with available retries can be ready to run.
+	job := c.jobChain.Jobs[jobId]
+	switch job.State {
+	case proto.STATE_PENDING:
+	case proto.STATE_STOPPED:
```
daniel-nichter: `case proto.STATE_PENDING, proto.STATE_STOPPED`. I suspect this doesn't work because Go switch don't fall through, so this should fail for `proto.STATE_PENDING`.

job-runner/chain/chain.go
```diff
@@ -15,19 +15,29 @@ import (
 
 // chain represents a job chain and some meta information about it.
 type chain struct {
```
daniel-nichter: I think it'd help to make this `Chain` so we can see all it's methods from a high level (via godoc).

job-runner/chain/chain.go
```diff
@@ -38,10 +48,44 @@ func NewChain(jc *proto.JobChain) *chain {
...
+		totalTries:     make(map[string]uint),
+		totalTriesMux:  &sync.RWMutex{},
+		latestRunTries: make(map[string]uint),
+		latestTriesMux: &sync.RWMutex{},
+	}
+}
+
+// ResumeChain takes a SuspendedJobChain proto (from the RM) and turns it into a
+// Chain that the JR can use.
+func ResumeChain(sjc *proto.SuspendedJobChain) *chain {
```
daniel-nichter: All this logic is called in only 1 place with the goal of making a normal chain: `MakeFromSJC`. I think it makes more sense directly in that method than standing alone here. This code is basically just factory code to build a chain.

job-runner/chain/chain.go
```diff
@@ -97,103 +135,197 @@ func (c *chain) NextJobs(jobId string) proto.Jobs {
 }
 
 // JobIsReady returns whether or not a job is ready to run. A job is considered
-// ready to run if all of its previous jobs are complete. If any previous jobs
-// are not complete, the job is not ready to run.
+// ready to run if it is Pending or Stopped with some retry attempts remaining,
+// and all of its previous jobs are complete. If any previous jobs are not
+// complete, the job is not ready to run.
 func (c *chain) JobIsReady(jobId string) bool {
```
daniel-nichter: Renaming to `IsRunnable(jobId string) bool` will align with `RunnableJobs()`.

daniel-nichter: If we settle on "runnable" then let's be sure not to use "ready" anywhere, i.e. be consistent with the "runnable job" term and define what that means.

job-runner/chain/chain.go
daniel-nichter: Better yet, see comment above ^

job-runner/chain/chain.go
```diff
@@ -97,103 +135,197 @@ func (c *chain) NextJobs(jobId string) proto.Jobs {
...
+		// If there is at least one job that is not complete, the whole chain
+		// is not complete. The chain could still be done, though, so we aren't
+		// ready to return yet.
 		complete = false
 	}
 
-	// For each pending job, check to see if all of its previous jobs
-	// completed. If they did, there's no reason the pending job can't run.
+	// For each pending job, check to see if it's ready to be run.
 	for _, job := range pendingJobs {
```
daniel-nichter: I think this loop is superfluous. Instead, do the check in the switch statement, no need to build a list:
```
        case proto.STATE_PENDING, proto.STATE_STOPPED:
            // Treat stopped jobs as pending jobs because they may be retried
            // (as in a resumed job chain)
            !if c.jobIsReady(job.Id) {
                return false, false
            }
```

job-runner/chain/chain.go
```diff
@@ -97,103 +135,197 @@ func (c *chain) NextJobs(jobId string) proto.Jobs {
...
+			break
 		}
 	}
+	return done, complete
+}
+
+func (c *chain) SequenceStartJob(jobId string) proto.Job {
+	c.jcMux.RLock()
+	defer c.jcMux.RUnlock()
+	return c.jobChain.Jobs[c.jobChain.Jobs[jobId].SequenceId]
```
daniel-nichter: Does this need to be guarded? The sequence start job should never change.

job-runner/chain/chain.go
```diff
@@ -97,103 +135,197 @@ func (c *chain) NextJobs(jobId string) proto.Jobs {
...
+}
+
+func (c *chain) IsSequenceStartJob(jobId string) bool {
+	c.jcMux.RLock()
+	defer c.jcMux.RUnlock()
+	return (jobId == c.jobChain.Jobs[jobId].SequenceId)
```
daniel-nichter: The `( )` aren't needed.

daniel-nichter: And same as above: does this need to be guarded?

job-runner/chain/reaper.go
```diff
@@ -0,0 +1,435 @@
...
+}
+
+// Create a new ReaperFactory.
+func NewReaperFactory(chain *chain, chainRepo Repo, rmc rm.Client) ReaperFactory {
+	return NewReaperFactoryWithRetries(chain, chainRepo, rmc, FINALIZE_TRIES, FINALIZE_RETRY_WAIT)
+}
+
+// Create a new ReaperFactory with custom settings for retrying RMClient calls.
+// Used in testing to decrease default retry wait duration.
+func NewReaperFactoryWithRetries(chain *chain, chainRepo Repo, rmc rm.Client, rmTries int, rmRetryWait time.Duration) ReaperFactory {
```
daniel-nichter: I don't see why this is necessary? It strikes me as an anti-pattern: two factories producing the same thing. Difficult to explain/show alternative here. Can discuss in person.

job-runner/chain/reaper.go
```diff
@@ -0,0 +1,435 @@
...
+
+const (
+	// Number of times to attempt sending chain state / SJC to RM.
+	FINALIZE_TRIES = 5
+	// Time to wait between the above tries.
+	FINALIZE_RETRY_WAIT = 500 * time.Millisecond
+)
+
+// A JobReaper handles jobs and chains that have finished running.
+type JobReaper interface {
```
daniel-nichter: Let's document more here: why we have varying implementations of reapers. Or document it elsewhere. It's an important design decision.

job-runner/chain/traverser.go
```diff
@@ -1,15 +1,27 @@
-// Copyright 2017-2018, Square, Inc.
+// Copyright 2018, Square, Inc.
```
daniel-nichter: Typo?

job-runner/chain/traverser.go
```diff
@@ -21,7 +33,7 @@ type Traverser interface {
 	// are no more jobs to run, or until the Stop method is called on the traverser.
 	//
 	// It returns an error if it fails to start.
-	Run() error
+	Run()
```
daniel-nichter: Typo? Docs say it returns an error but the `error` return was removed.

job-runner/chain/traverser.go
```diff
@@ -81,209 +107,159 @@ func (f *traverserFactory) Make(jobChain proto.JobChain) (Traverser, error) {
...
+	stopMux *sync.RWMutex
+
+	chain      *chain
+	chainRepo  Repo // stores all currently running chains
+	rf         runner.Factory
+	runnerRepo runner.Repo // stores actively running jobs
+	rmc        rm.Client
+	logger     *log.Entry
+
+	StopTimeout time.Duration // Time to wait for runners to return after being stopped.
```
daniel-nichter: Why is only this field exported?

job-runner/chain/traverser.go
```diff
@@ -81,209 +107,159 @@ func (f *traverserFactory) Make(jobChain proto.JobChain) (Traverser, error) {
...
+		t.reaper.Reap(doneJob)
+		t.stopMux.Unlock()
 	}
 
-	return nil
+	// Indicate that the chain is done running. Stops waitForShutdown() listener.
+	select {
+	case <-t.doneChan:
+	default:
+		close(t.doneChan)
```
daniel-nichter: I don't understand this. The only place `doneChan` is closed is here, but this also checks if `doneChan` is closed. The code comment doesn't clarify either. So I cannot tell what/how/when `doneChan` would be closed.

job-runner/chain/traverser.go
```diff
@@ -81,209 +107,159 @@ func (f *traverserFactory) Make(jobChain proto.JobChain) (Traverser, error) {
...
+// Stop stops a running chain by stopping all its running jobs.
 func (t *traverser) Stop() error {
+	// LOCK - can't close these chans while a job is being Reaped / initial
+	// jobs are being sent to runJobChan.
+	t.stopMux.Lock()
+
+	// Don't do anything if traverser has already been stopped.
+	select {
+	case <-t.stopChan:
+		return nil
```
daniel-nichter: `stopMux` is not unlocked, so a 2nd call to Stop will deadlock. Should almost always used `defer <unlock>` after locking unless not possible.

job-runner/server/server.go
```diff
@@ -61,47 +62,63 @@ func (s *Server) API() *api.API {
...
 
-func loadConfig(appCtx *app.Context) error {
+// Catch TERM and INT signals to gracefully shut down the Job Runner
+func (s *Server) waitForShutdown() {
+	s.sigChan = make(chan os.Signal, 1)
+	signal.Notify(s.sigChan, syscall.SIGINT, syscall.SIGTERM)
+
+	sig := <-s.sigChan
+	if sig == nil {
+		// channel was closed
```
daniel-nichter: A select is more idiomatic:
```go
select {
case <-sig:
// handle it
default:
// chan closed, ignore
}
```

daniel-nichter: But why would the channel ever be closed?

proto/const.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2017, Square, Inc.
+// Copyright 2017-2018, Square, Inc.
```
daniel-nichter: Typo?

proto/s2s.go
```diff
@@ -51,6 +51,22 @@ type Request struct {
...
+	RequestId string    `json:"requestId"`
+	JobChain  *JobChain `json:"jobChain"`
+	// job.Id -> total number of times a job has been tried
+	TotalJobTries map[string]uint `json:"totalJobTries"`
+	// job.Id -> number of times a job was tried the last time it was run
+	LastRunJobTries map[string]uint `json:"lastRunJobTries"`
+	// Id of 1st job in sequence -> number of times sequence has been tried
+	SequenceTries map[string]uint `json:"sequenceTries"`
+	// number of jobs that have been run (each job counts only once, even if retried)
+	NumJobsRun uint `json:"numJobsRun"`
```
daniel-nichter: Does this need to be stored, or can it be recalculated when the SJC is resumed by counting completed jobs?

proto/s2s.go
```diff
@@ -51,6 +51,22 @@ type Request struct {
...
+// SuspendedJobChain (SJC) represents the data required to reconstruct and resume a
+// running job chain in the Job Runner.
+type SuspendedJobChain struct {
+	// request the SJC is for - unique identifier
+	RequestId string    `json:"requestId"`
+	JobChain  *JobChain `json:"jobChain"`
+	// job.Id -> total number of times a job has been tried
+	TotalJobTries map[string]uint `json:"totalJobTries"`
+	// job.Id -> number of times a job was tried the last time it was run
+	LastRunJobTries map[string]uint `json:"lastRunJobTries"`
```
daniel-nichter: From comments it's unclear why we have 2 job counts. I presume this has to do with sequence retries? I.e. Job has been tried 5 times, and last time (for last seq. retry) it was tried 2 times (which implies something like seq. try 1 + 3 job tries and seq. try 2 + 2 job tries (1 remaining), if it's retry=3).

proto/const.go
```diff
@@ -21,24 +21,28 @@ const (
 	STATE_FAIL    // failed due to error or non-zero exit
 	STATE_TIMEOUT // timeout
 	STATE_STOPPED // stopped by user
+
+	STATE_SUSPENDED // suspended; doesn't apply to jobs
```
daniel-nichter: What does "doesn't apply to jobs" mean? The other states do apply but this one doesn't?

job-runner/runner/factory.go
```diff
@@ -26,7 +26,12 @@ func NewFactory(jf job.Factory, rmc rm.Client) Factory {
 	}
 }
 
-func (f *factory) Make(pJob proto.Job, requestId string, prevTryNo uint, sequenceRetry uint) (Runner, error) {
+// Make a runner for a new job.
+func (f *factory) Make(pJob proto.Job, requestId string, prevTryNo uint, triesToSkip uint, sequenceRetry uint) (Runner, error) {
+	// Remove a number of retries from the job. Useful when resuming a
```
daniel-nichter: "Useful" or necessary?

job-runner/runner/runner.go
```diff
@@ -133,9 +141,19 @@ TRY_LOOP:
...
+				return r.rmc.CreateJL(r.reqId, jl)
+			},
+			func(err error) {
+				tryLogger.Errorf("problem sending job log (%#v) to the RM: %s. Retrying...", jl, err)
+			},
+		)
+		if err != nil {
+			tryLogger.Errorf("failed to send job log (%#v) to Request Manager", jl)
 		}
+		tryLogger.Infof("successfully sent job log (%#v) to Request Manager", jl)
```
daniel-nichter: I wouldn't log this. We only care when it fail to send; we expect it to send ok.

job-runner/runner/runner.go
```diff
@@ -133,9 +141,19 @@ TRY_LOOP:
...
 		}
-		if err := r.rmc.CreateJL(r.reqId, jl); err != nil {
-			tryLogger.Errorf("problem sending job log (%#v) to the RM: %s", jl, err)
+		// Send the JL to the RM.
+		err := retry.Do(JOB_LOG_TRIES, JOB_LOG_RETRY_WAIT,
+			func() error {
+				return r.rmc.CreateJL(r.reqId, jl)
+			},
+			func(err error) {
+				tryLogger.Errorf("problem sending job log (%#v) to the RM: %s. Retrying...", jl, err)
```
daniel-nichter: I wouldn't log this either. We don't care about transient failures. Logging the final error is good enough.

job-runner/runner/factory.go
```diff
@@ -11,7 +11,7 @@ import (
 // A Factory takes a proto.Job, creates a corresponding job.Job interface for
 // it, and passes both to NewRunner to make a Runner.
 type Factory interface {
-	Make(job proto.Job, requestId string, prevTryNo uint, sequenceRetry uint) (Runner, error)
+	Make(job proto.Job, requestId string, prevTryNo uint, triesToSkip uint, sequenceRetry uint) (Runner, error)
```
daniel-nichter: The var name `triesToSkip` is a little misleading. We never skip tries. In this case, we've already done those tries. Not sure what a better name is, but we need to think of one.

job-runner/chain/traverser.go
```diff
@@ -81,209 +107,159 @@ func (f *traverserFactory) Make(jobChain proto.JobChain) (Traverser, error) {
...
+	default:
+	}
+
 	t.logger.Infof("stopping traverser and all jobs")
 
-	// Stop the traverser (i.e., stop running new jobs).
-	close(t.stopChan)
+	t.reaper = t.reaperFactory.MakeStopped()
```
daniel-nichter: This needs a code comment. It's an important change and not obvious why we do it.

job-runner/chain/chain.go
felixplajer: For testing - proto.Jobs lets you sort the list of jobs to compare it to an expected list. Other than that, proto.Jobs is the same as []proto.Job.

job-runner/chain/chain.go
felixplajer: I think we already talked about this kinda thing in the context of the sequenceTries map:
SequenceId won't change, but we need a lock for reading from the Jobs map - it gets written to when changing job state, and concurrent map reads + writes aren't safe even if they're for different keys.

job-runner/chain/chain.go
felixplajer: same thing as in SequenceStartJob - it does need the lock.

job-runner/chain/traverser.go
```diff
@@ -81,209 +110,169 @@ func (f *traverserFactory) Make(jobChain proto.JobChain) (Traverser, error) {
...
+	t.reaper = t.reaperFactory.MakeRunning(t.runningJobs)
+	go t.reaper.Run()
+
+	<-t.doneChan
 }
 
-// Stop stops the traverser if it's running.
+// Stop stops a running chain by stopping all its running jobs.
 func (t *traverser) Stop() error {
+	// Don't do anything if traverser has already been stopped.
```
daniel-nichter: This isn't thread-safe. Pretend 3 calls to Stop at same time. All three will hit default case on line 259, make a new reaper, and close the chan--2 of them panic on close. Need a mutex and a t.stopped "flag" to serialize access and guarantee only-once execution of this func. (Or the the atomic.Int trick, but haven't found a good example yet.)

job-runner/chain/traverser.go
```diff
@@ -320,186 +309,190 @@ func (t *traverser) Status() (proto.JobChainStatus, error) {
...
 
-// runJobs loops on the runJobChannel, and for each job that comes through the
-// channel, it creates a job.Job interface and runs it in a goroutine. If there
-// are any errors creating the job.Job it creates a JL that contains the error
-// and attaches it to the job. When it's done, it sends the job out through the
-// doneJobChan.
+// runJobs loops on the runJobChan, and runs each job that comes through the
+// channel. When the job is done, it sends the job out through the doneJobChan.
+// Once runJobChan is closed, it waits for any running jobs to finish and
+// then closes doneJobChan.
```
daniel-nichter: This comment is no longer true.

job-runner/chain/traverser.go
```diff
@@ -81,209 +110,169 @@ func (f *traverserFactory) Make(jobChain proto.JobChain) (Traverser, error) {
...
-				t.retrySequence(doneJob) // re-enqueue first job of failed sequence
-				continue JOB_REAPER
-			}
-
-			// If the failed job is not part of a retryable sequence, then ignore
-			// subsequent jobs. For example, if job B of A -> B -> C  fails, then C is
-			// not ran.
-			jLogger.Warn("job did not complete successfully")
-			continue JOB_REAPER
+	// Suspend running job chain if Job Runner shuts down.
```
daniel-nichter: Can move this block to end of func, replacing the single `<-t.doneChan` line with this block sans the goroutine. This block is the complete "when does this func stop" code.

job-runner/chain/reaper.go
```diff
@@ -0,0 +1,589 @@
...
+// - job completed: prepared subsequent jobs and enqueue if runnable.
+func (r *RunningChainReaper) Run() {
+REAPER:
+	for {
+		select {
+		case job := <-r.doneJobChan:
+			delete(r.running, job.Id)
+			r.Reap(job)
+		case <-r.stopChan:
+			close(r.doneStopping)
```
daniel-nichter: Need to close this after r.Finalize()?

job-runner/chain/reaper.go
```diff
@@ -0,0 +1,589 @@
...
+	for _, nextJob := range r.chain.NextJobs(job.Id) {
+		nextJLogger := jLogger.WithFields(log.Fields{"next_job_id": nextJob.Id})
+
+		for k, v := range job.Data {
+			nextJob.Data[k] = v
+		}
+		r.chainRepo.Set(r.chain)
+
+		// Enqueue any runnable child jobs.
+		if !r.chain.IsRunnable(nextJob.Id) {
```
daniel-nichter: Do this first, i.e. don't copy data and call chainRepo.Set then skip the job because it's not runnable.

job-runner/chain/reaper.go
```diff
@@ -0,0 +1,589 @@
...
+// Run reaps jobs when they finish running. For each job reaped, if it's...
+// - completed: prepare subsequent jobs (copy jobData).
+// - failed:    prepare a sequence retry.
+// - stopped:   do nothing (job will be retried when chain is resumed).
+func (r *SuspendedChainReaper) Run() {
+	// Wait at most 10 seconds for all running jobs to finish.
+	timeoutChan := make(chan struct{})
+	go func() {
+		time.Sleep(r.stopTimeout)
+		close(timeoutChan)
```
daniel-nichter: Does this block work? It seems not to do what it intends.

daniel-nichter: Oh, I see what it's doing. Instead of this block, you want `case <-time.After(r.stopTimeout):` instead of `case <-timeoutChan:`.  Also, a comment somewhere that explains that this (and the Stopped) reaper only run for a short time to give in-flight jobs time to finish and be reaped. The traverser doesn't explicitly stop this reaper; it needs to stop itself.

Explained that way, I think it makes more sense for caller (traverser) to wait and stop. This internal timeout is kind of magical. Reaper.Run() should run until down and return, or until its Stop method is called. So traverser should swap in this reaper, run it and either wait for it to return or timeout and call its Stop method.

job-runner/chain/reaper.go
```diff
@@ -0,0 +1,589 @@
...
+		case job := <-r.doneJobChan:
+			delete(r.running, job.Id)
+			r.Reap(job)
+		case <-r.stopChan:
+			close(r.doneStopping)
+			return
+		default:
+			// If there were no jobs done running, and there are no jobs currently
+			// running, then there must be no more jobs to run - the chain is done.
+			if len(r.running) == 0 {
```
daniel-nichter: This creates an infinite, uncontrolled loop at times. For example, if we start a long-running job, this `for` loop as fast as possible, hitting the default case. That's too inefficient. Need a more controlled way to know when done.

job-runner/chain/reaper.go
```diff
@@ -0,0 +1,589 @@
...
+		default:
+			// All jobs have finished running - no more jobs to reap.
+			if len(r.running) == 0 {
+				break REAPER
+			}
+		}
+	}
+
+	r.Finalize()
+	close(r.doneChan)
```
daniel-nichter: Reapers should not be aware of this: doneChan is what stops the `traverser.runJobs` gorountine, but that's owned by the traverser. The traverser and its runJobs is at a higher/different level than the reapers, so the reapers shouldn't be aware of this.

This jives with my earlier comment that we remove the magical timeout from {Stopped|Suspened}Reapder.Run, making the stop or suspend sequence this, all in traverser:

1. Stop RunningReaper
2. Swap in {Stopped|Suspended}Reaper
3. Wait for ^.Run() to return; or timeout and call ^.Stop()
4. defer close(t.doneChan) to stop `traverser.runJobs`

Let me spec several things because this is all related code...

job-runner/chain/traverser.go
```diff
@@ -81,209 +110,169 @@ func (f *traverserFactory) Make(jobChain proto.JobChain) (Traverser, error) {
...
-	// job.Id -> number of tries
-	jobTries map[string]uint
+// -------------------------------------------------------------------------- //
 
-	// mutex to access jobTries list
-	jtMux *sync.RWMutex
+type traverser struct {
+	reaperFactory ReaperFactory
+	reaper        JobReaper
+	runningJobs   map[string]struct{}
```
daniel-nichter: I understand the usage but I think the runnerRepo needs to be the single source of truth for what's running.

job-runner/chain/reaper.go
felixplajer: The idea behind this was to copy the data of all parent jobs to each child job, not just the data from the last parent job to finish.

daniel-nichter: Code comment :-)

job-runner/chain/reaper.go
```diff
@@ -0,0 +1,642 @@
...
+func (r *RunningChainReaper) Finalize(complete bool) {
+	if complete {
+		r.logger.Infof("chain is done, all jobs finished successfully")
+		r.chain.SetState(proto.STATE_COMPLETE)
+	} else {
+		r.logger.Warn("chain is done, some jobs failed")
+		r.chain.SetState(proto.STATE_FAIL)
+	}
+	r.chainRepo.Set(r.chain)
+	if r.sendFinalState() {
```
daniel-nichter: This isn't idiomatic Go. Was confusing at first until I read the func. More idiomatic:
```
if err := r.sendFinalState(); err != nil {
    // log
}
r.chainRepo.Remove(r.chain.RequestId())
```
I presume we should always remove chain from repo, even if final state not sent? If not removed from repo, how/when will it ever be removed?

job-runner/chain/reaper.go
```diff
@@ -0,0 +1,642 @@
...
+		case job := <-r.doneJobChan:
+			r.Reap(job)
+
+			// If runnerRepo is consistently empty, there are no more running jobs
+			// and thus no more jobs to be reaped. Break out of the reaper loop.
+			if r.runnerRepo.Count() == 0 {
+				// runnerRepo might be empty very briefly after a job is sent to
+				// runJobChan but before its runner is created, so wait a while
+				// and double check.
+				time.Sleep(500 + time.Millisecond)
```
daniel-nichter: Typo: `500 *  time.Millisecond)`

job-runner/chain/reaper.go
```diff
@@ -0,0 +1,642 @@
...
+				if r.runnerRepo.Count() == 0 {
```
daniel-nichter: This logic is duplicated in `done()` so let that be single source of truth. Then here:
```
if r.done() {
    break REAPER
}
```

job-runner/chain/traverser.go
```diff
@@ -320,186 +358,216 @@ func (t *traverser) Status() (proto.JobChainStatus, error) {
...
+	// Don't do anything if the traverser has already finished running, or has
+	// already been stopped or suspended.
+	t.stopMux.Lock()
+	select {
+	case <-t.doneChan:
+		t.stopMux.Unlock()
+		return
+	default:
 	}
+	if t.stopped || t.suspended {
```
daniel-nichter: I think we only need to lock and check these bool vars, don't need to check chan.

job-runner/chain/traverser.go
```diff
@@ -81,209 +115,213 @@ func (f *traverserFactory) Make(jobChain proto.JobChain) (Traverser, error) {
...
 		}
-
-		// Update chain repo for jobs next jobs just started ^
-		t.chainRepo.Set(t.chain)
+		t.stopMux.Unlock()
+	case <-t.shutdownChan:
+		// Job Runner is shutting down. Stop the running reaper and suspend
+		// the job chain, to be resumed later.
+		t.shutdown()
+		close(t.runJobChan) // can close because we've stopped the running reaper
```
daniel-nichter: Remove this to avoid double-close, and instead...

job-runner/chain/traverser.go
```diff
@@ -81,209 +115,213 @@ func (f *traverserFactory) Make(jobChain proto.JobChain) (Traverser, error) {
...
+		close(runningReaperChan)
+	}()
 
-			t.runJobChan <- nextJ // add the job to the run queue
+	// Wait for running reaper to be done or traverser to be shut down.
+	select {
+	case <-runningReaperChan:
+		// Running reaper is done - nothing else will send to runJobChan.
+		// Close runJobChan to stop runJobs loop.
+		close(t.runJobChan)
```
daniel-nichter: Remove this to avoid double-close, and instead...

job-runner/chain/traverser.go
```diff
@@ -81,209 +115,213 @@ func (f *traverserFactory) Make(jobChain proto.JobChain) (Traverser, error) {
...
-			for k, v := range doneJob.Data {
-				nextJ.Data[k] = v
-			}
+	// Start a goroutine to reap done jobs. The Reaper consumes from doneJobChan
+	// and sends the next jobs to be run to runJobChan.
+	runningReaperChan := make(chan struct{})
+	t.reaper = t.reaperFactory.MakeRunning()
+	go func() {
+		t.reaper.Run()
+		close(runningReaperChan)
```
daniel-nichter: ...instead:
```
go func() {
    // defer is LIFO, so when the running reaper stops for any reason, first close
    // runJobChan to stop the runJobs goroutine, then close runningReaperChan
    // to unblock the select waits below and return from this func (Run).
    defer close(runningReaperChan)
    defer close(t.runJobChan)
    t.reaper.Run()
}()
```

job-runner/chain/reaper.go
felixplajer: Yeah, you're right - we should always remove the chain from the repo. Not removing it if there was a problem sending the state was a leftover from the old traverser code, but it doesn't really make sense. We already log the error in sendFinalState, so I'll change it to not return anything.

job-runner/chain/chain.go
```diff
@@ -14,34 +14,61 @@ import (
...
+	sequenceTries     map[string]uint // Number of sequence retries attempted so far
+	latestRunJobTries map[string]uint // job.Id -> number of times tried within the latest time it was run (i.e. within the latest sequence try)
+	totalJobTries     map[string]uint // job.Id -> total number of times tried
 }
 
-// NewChain takes a JobChain proto (from the RM) and turns it into a Chain that
-// the JR can use.
-func NewChain(jc *proto.JobChain) *chain {
-	sequenceRetryCount := make(map[string]uint)
+// NewChain takes a JobChain proto and maps of sequence + jobs trues, and turns them
```
michaelfinch: tries*

job-runner/chain/chain.go
```diff
@@ -14,34 +14,61 @@ import (
...
+		case proto.STATE_COMPLETE:
+			// Valid state, job is done running.
+			numJobsRun += 1
+		case proto.STATE_FAIL:
+			// Valid state, job is done running.
+			numJobsRun += 1
+		default:
+			// Job isn't pending, failed, or complete. For a new / suspended chain,
+			// these are the only valid states (no jobs can be running before the
+			// chain is started or resumed). Set jobs with other states to pending.
```
michaelfinch: So jobs with STATE_TIMEOUT and STATE_STOPPED will both be set to STATE_PENDING, right? Just want to make sure that's intended

job-runner/chain/chain.go
```diff
@@ -96,108 +117,167 @@ func (c *chain) NextJobs(jobId string) proto.Jobs {
...
-// ready to run if all of its previous jobs are complete. If any previous jobs
-// are not complete, the job is not ready to run.
-func (c *chain) JobIsReady(jobId string) bool {
-	c.jcMux.RLock()
-	defer c.jcMux.RUnlock()
-	for _, job := range c.previousJobs(jobId) {
-		if job.State != proto.STATE_COMPLETE {
-			return false
+// IsRunnable returns whether or not a job is runnable. A job is considered
+// runnable if it is Pending or Stopped with some retry attempts remaining,
```
michaelfinch: When would a Stopped job be runnable? When we make a NewChain, don't all Stopped jobs get their state changed to Pending?

job-runner/chain/chain.go
```diff
@@ -96,108 +117,167 @@ func (c *chain) NextJobs(jobId string) proto.Jobs {
...
-// IsDone returns two booleans - the first one indicates whether or not the
-// chain is done, and the second one indicates whether or not the chain is
-// complete.
+// IsDoneRunning returns two booleans - the first indicates whether the chain
+// is done, and the second indicates whether the chain is complete.
 //
-// A chain is done running if there are no more jobs in it that can run. This
-// can happen if all of the jobs in the chain or complete, or if some or all
-// of the jobs in the chain failed.
+// A chain is done running if there are no more jobs in it that can be run or
```
michaelfinch: "A chain is done running if there are no jobs in it running and there are no more jobs in it that can be run." is more clear to me

job-runner/chain/chain.go
```diff
@@ -292,66 +371,136 @@ func (c *chain) SetJobState(jobId string, state byte) {
...
+	c.runningMux.RLock()
+	running := c.running
+	numJobsRun := c.numJobsRun
+	c.runningMux.RUnlock()
 
 	m := chainJSON{
-		JobChain: c.jobChain,
-		Running:  c.running,
-		N:        c.n,
+		SuspendedJobChain: c.ToSuspended(),
```
michaelfinch: I'm a little confused as to why whenever you marshal a job chain you will get a suspended job chain. Maybe a comment would help

job-runner/chain/chain.go
```diff
@@ -292,66 +371,136 @@ func (c *chain) SetJobState(jobId string, state byte) {
...
 
 	return nil
 }
 
 // -------------------------------------------------------------------------- //
 
+// isRunnable returns whether or not a job is runnable. A job is considered
+// runnable if it is Pending or Stopped with some retry attempts remaining,
+// and all of its previous jobs are complete. If any previous jobs are not
+// complete, the job is not unnable.
```
michaelfinch: *runnable

job-runner/chain/chain.go
```diff
@@ -292,66 +371,136 @@ func (c *chain) SetJobState(jobId string, state byte) {
...
+// Doesn't lock jobsMux, so suitable to call within other Chain methods lock jobsMux.
```
michaelfinch: *other Chain methods that lock

Is it only safe to call this method if you've already locked jobsMux? If so, mention that

job-runner/chain/chain.go
```diff
@@ -14,34 +14,61 @@ import (
...
+// NewChain takes a JobChain proto and maps of sequence + jobs trues, and turns them
+// into a Chain that the JR can use.
+func NewChain(jc *proto.JobChain, sequenceTries map[string]uint, totalJobTries map[string]uint, latestRunJobTries map[string]uint) *Chain {
+	// Make sure all jobs have valid State + Data fields, and count the number of
+	// completed + failed jobs (the number of jobs that have finished running).
+	numJobsRun := uint(0)
 	for jobName, job := range jc.Jobs {
-		job.State = proto.STATE_PENDING
-		job.Data = map[string]interface{}{}
+		switch job.State {
```
michaelfinch: Is this logic tested?

job-runner/chain/chain.go
michaelfinch: Also, in a follow up PR, we should probably get rid of the chain.RedisRepo (we've never actually used it). It marshals the Chain struct when you add a Chain to the repo, and I'm not sure if turning the Chain into a suspended Chain every time that happens would have unintended consequences.

job-runner/chain/chain.go
felixplajer: Ah, good catch, actually Stopped jobs need to stay STATE_STOPPED so we get the number of job tries right later on. Changed so Stopped jobs stay Stopped and all other states (Running, Timeout) become Failed instead of Pending.

job-runner/chain/chain.go
felixplajer: b/c of the above^ change to keep Stopped jobs Stopped instead of Pending, this can stay as-is.

job-runner/chain/chain.go
felixplajer: This was just a sort of convenience, since chain.ToSuspended() already saves most of the things you need to know to recreate a Chain. I see how it's confusing to "suspend" the chain whenever you marshal it, though. I'll change this so it doesn't use an SJC or ToSuspended() anymore.

As for the chain repo, we can either get rid of it somewhen or adapt it as a way to recover job chains after Job Runner crashes.

job-runner/server/server.go
```diff
@@ -61,47 +61,59 @@ func (s *Server) API() *api.API {
...
-func loadConfig(appCtx *app.Context) error {
+// Catch TERM and INT signals to gracefully shut down the Job Runner
+func (s *Server) waitForShutdown() {
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+
+	<-sigChan
+
+	// API + traversers watch shutdownChan
+	close(s.shutdownChan)
```
BrianIp: This will send a signal to the API and traversers to begin the shutdown process, but does it wait for them to finish shutting down?

job-runner/chain/reaper.go
```diff
@@ -0,0 +1,630 @@
...
+	Run()
+
+	// Stop stops the JobReaper from reaping any more jobs. It blocks until
+	// Run() returns and the reaper can be safely switched out for another
+	// implementation.
+	Stop()
+}
+
+// --------------------------------------------------------------------------
+// JobReapers:
```
michaelfinch: Great comment

job-runner/chain/reaper.go
```diff
@@ -0,0 +1,630 @@
...
+	reaper
+	runJobChan chan proto.Job // enqueue next jobs to run here
+}
+
+// Run reaps jobs when they finish running. For each job reaped, if...
+// - chain is done: save final state + send to RM.
+// - job failed:    retry sequence if possible.
+// - job completed: prepared subsequent jobs and enqueue if runnable.
+func (r *RunningChainReaper) Run() {
+	defer close(r.doneChan)
```
michaelfinch: At first glance this looks dangerous to me because, in general, only senders should close channels, else you run the risk of the senders trying to send to a closed channel. Are we certain this is safe? Maybe a comment would help here

job-runner/chain/reaper.go
```diff
@@ -0,0 +1,630 @@
...
+	return
+}
+
+// reap takes a job that just finished running, saves its final state, and prepares
+// to continue running the chain (or recognizes that the chain is done running).
+//
+// If chain is done: save final state + stop running more jobs.
+// If job failed:    retry sequence if possible.
+// If job completed: prepared subsequent jobs and enqueue if runnable.
+func (r *RunningChainReaper) Reap(job proto.Job) {
```
michaelfinch: Does this method need to be public? I don't see it defined on the interface. Or is this just to make testing easier?

job-runner/chain/reaper.go
```diff
@@ -0,0 +1,630 @@
...
+// - stopped:   do nothing (job will be retried when chain is resumed).
+func (r *SuspendedChainReaper) Run() {
+	defer close(r.doneChan)
+
+	// Sleep a short time to prevent a race condition on the runner Repo. If a job
+	// was sent to runJobChan just before the chain was suspended, its runner might
+	// not have been created + added to the Repo yet. Wait so we can be sure all
+	// running jobs have runners in the Repo, so our checks to runnerRepo.Count
+	// will accurately reflect whether there are any running jobs left.
+	time.Sleep(runnerRepoWait)
```
michaelfinch: Would it be possible to not send jobs to runJobChan until their runners have been created and added to the Repo?

job-runner/chain/reaper.go
```diff
@@ -0,0 +1,630 @@
...
+		return
+	}
+
+	// Reap jobs until there are no jobs left running, or the reaper is stopped.
+REAPER:
+	for r.runnerRepo.Count() > 0 {
+		select {
+		case job := <-r.doneJobChan:
+			r.Reap(job)
+		case <-time.After(runnerRepoWait):
```
michaelfinch: Why do we need this case? Can't we just select on doneJobChan and stopChan? Or is there a case where runnerRepo.Count() will be greater than 0, but we'll never receive anything on doneJobChan?

job-runner/chain/reaper.go
```diff
@@ -0,0 +1,630 @@
...
+	// Process sequenceStartJob
+	for _, pJob := range r.chain.NextJobs(sequenceStartJob.Id) {
+		toVisit[pJob.Id] = pJob
+	}
+	visited[sequenceStartJob.Id] = sequenceStartJob
+
+PROCESS_TO_VISIT_LIST:
+	for len(toVisit) > 0 {
+
+	PROCESS_CURRENT_JOB:
```
michaelfinch: Maybe I'm missing something here, but once we get to the last job in the sequence, what's stopping this from continuing to look at that job's next jobs?

michaelfinch:
Overall, great work on this PR. It's dense and relatively complex, but you've organized your code and comments in a way that makes it pretty straightforward to follow.

job-runner/chain/reaper.go
felixplajer: It's just for testing.

job-runner/chain/reaper.go
felixplajer: It's *possible*, but I'd rather not. It would mean moving logic about running the jobs into the reaper, which should only be dealing with jobs after they're done running.

Not sure if the comment makes it clear, but sending the job to runJobChan is what causes the runner to be created (the loop in the traverser that reads from runJobChan creates the runners) - it just doesn't happen instantly, hence the wait.

job-runner/chain/reaper.go
felixplajer: Yeah, we need this case - it is possible for runnerRepo.Count to be > 0 even though there's no jobs left running (and so nothing will get sent to doneJobChan). The runner gets removed from the repo right after the job gets sent to doneJobChan in traverser.runJobs(). Reaping a job is pretty quick for suspended/stopped chains, so it's possible for the reaper to receive the job, reap it, and go back to check the `r.runnerRepo.Count() > 0` loop condition before traverser.runJobs() gets a chance to remove the runner from the repo. If that happens, we need this case or we'd be stuck in the loop.

I'll add a comment to that ^ effect.

job-runner/server/server.go
felixplajer: Hmm, you're right, this doesn't really wait for anything to shut down - just catches the SIGINT/SIGTERM, which stops the JR from getting shut down until it eventually gets killed (by which point, ideally, the traversers have had a chance to shut down). It's better, though, to have the JR shut down on its own once the traversers + API are all done shutting down - I'll add code for that.

job-runner/chain/reaper.go
michaelfinch: Ack, thanks for tracking that down. Agreed, let's fix it in a future PR.

job-runner/chain/reaper.go
michaelfinch: After reviewing the traverser this makes sense now. What are the implications if the sleep does not prevent the race condition, due to something like a GC pause, network latency, etc. (as Bill says, "One in a million is next Tuesday")?

felixplajer: In that case, the reaper would return right away (per the `if runnerRepo.Count == 0` at line 309), without sending the final chain state or suspended job chain back to the RM. That's not great - the RM would think the request is still Running, even though the JR running it has shut down (basically, the same thing that happens if you shut down the current version of the JR while it's running a job chain).

We could increase the length of the sleep to make that less likely, but it's not a perfect solution.

michaelfinch: It's fine for now, but let's make a ticket to revisit and try to improve this later

PR #60, opened 2018-10-18
Remove references to old package name

bwester:
The "./jobs" package was moved to "./dev/jobs" in commit c38fdfd1. This
changes package references to use the new name and removes the old
symlink to get rid of the old package name completely.


michaelfinch:
Out of curiosity, how did you notice this?

bwester:
Now, I'm pretty sure that nothing has changed here or in downstream packages recently, and I don't remember seeing this error out of `godep` before. So I don't know where its coming from, but it's reproducible for me, and easily fixable too.

PR #61, opened 2018-10-19
Add placeholder jobs implementation

bwester:
This commit reverts commit fbb079de in #60.  In that PR, I misunderstood
how the "github.com/square/spincycle/jobs" package was being used.
SpinCycle is designed to have customizable actions, and the way to
customize them is to replace the contents of that package with an
implementation tailored for a domain.  Having the "./jobs" package as a
symlink is supposed to make it easy to replace.  In practice, it breaks
Go vendoring tools.

This commit replaces the symlink with a real "./jobs" package containing
a dummy job factory implementation.  To ensure that it is not
accidentally used in production, the factory panics.  To assist users
and future contributors, the package documents its intended use: to be
replaced with a real implementation.

### DIFF
--- a/dev/run-spincycle
+++ b/dev/run-spincycle
@@ -197,6 +197,15 @@ create_db() {
 }
 
 build() {
+  if [[ ! -L "$REPO_ROOT_DIR/jobs" ]]; then
+    echo "Installing development sandbox jobs..."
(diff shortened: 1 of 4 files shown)


bwester:
Thanks! (And I will take this commit to indicate I'm actually on the right track this time.)

PR #63, opened 2018-10-29
Request ACLs

daniel-nichter:
* Add request ACLs: auth pkg and grapher spec changes
* Add `Auth` section to config
* Refactor and document server/
* Expand app.Context and document
* Rename grapher "config" to "specs" in some places ("specs" is the normal term)
* Simplify `api.NewAPI`
* Set `Request.User` from auth plugin then override with `SetUsername` hook if defined
* Document hooks
* Remove default `SetUsername` hook (that set user = "admin")

### DIFF
--- a/request-manager/api/api.go
+++ b/request-manager/api/api.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/square/spincycle/proto"
 	"github.com/square/spincycle/request-manager/app"
+	"github.com/square/spincycle/request-manager/auth"
 	"github.com/square/spincycle/request-manager/db"
 	"github.com/square/spincycle/request-manager/joblog"
 	"github.com/square/spincycle/request-manager/request"
@@ -76,21 +77,15 @@ func NewAPI(appCtx app.Context, rm request.Manager, jls joblog.Store, stat statu
 	api.echo.Use(middleware.Recover())
 	api.echo.Use(middleware.Logger())
 
-	// Auth hook
+	// Auth plugin: authenticate user or app
 	api.echo.Use((func(next echo.HandlerFunc) echo.HandlerFunc {
 		return func(c echo.Context) error {
-			if appCtx.Hooks.Auth == nil {
-				return next(c) // no auth
-			}
-			ok, err := appCtx.Hooks.Auth(c.Request())
+			caller, err := appCtx.Plugins.Auth.Authenticate(c.Request())
 			if err != nil {
-				return err
+				return echo.NewHTTPError(http.StatusUnauthorized, err.Error())
 			}
-			if !ok {
-				//c.Response().Header().Set(echo.HeaderWWWAuthenticate, basic+" realm="+realm)
-				return echo.ErrUnauthorized // 401
-			}
-			return next(c) // auth OK
+			c.Set("caller", caller)
+			return next(c) // authenticed
 		}
 	}))
 
@@ -124,8 +119,6 @@ func (api *API) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	api.echo.ServeHTTP(w, r)
 }
 
-// ============================== CONTROLLERS ============================== //
-
 // POST <API_ROOT>/requests
 // Create a new request and start it.
 func (api *API) createRequestHandler(c echo.Context) error {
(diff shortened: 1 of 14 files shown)


request-manager/api/api.go
```diff
@@ -76,21 +76,16 @@ func NewAPI(appCtx app.Context, rm request.Manager, jls joblog.Store, stat statu
 	api.echo.Use(middleware.Recover())
 	api.echo.Use(middleware.Logger())
 
-	// Auth hook
+	// Auth plugin: authenticate caller. This is called before every route.
```
felixplajer: Do we need to keep the SetUsername hook (it's right below this)? It seems like it'll be doing the same thing as Auth.Authenticate - returning the caller. AFAIK it's used just for setting the Request.User field when creating a request - I think it'd make sense to have that set to the caller returned by Auth.

daniel-nichter: Yeah, thought about this. Let's leave it for now so that it's an easier way to set `proto.Request.User`. Else, users will have to implement a real object that implements the `auth.Plugin` interface just to set this. For example, until we enable auth, we can leave our existing hook in place.

request-manager/auth/auth.go
```diff
@@ -0,0 +1,154 @@
...
+REQUEST_ACLS:
+	for _, acl := range acls {
+		for _, crole := range caller.Roles {
+			// Does caller have this role?
+			if crole != acl.Role {
+				continue // no
+			}
+			rolesMatch = true // yes
+
+			// Is it an admin role? If yes, then allow all ops.
```
felixplajer: I was confused why you had both this check and the previous `isAdmin(caller)`, until I remembered there are some roles that are admins for *any* request, and also each request can have its own admin roles (giving access to all ops for that request). Maybe there could be a comment explaining this distinction?

request-manager/app/app.go
```diff
@@ -54,9 +96,16 @@ func Defaults() Context {
...
+		Plugins: Plugins{
+			Auth: auth.AllowAll{},
+		},
 	}
 }
 
+// LoadConfig is the default LoadConfig hook. It loads the config file specified
+// on the command line, or from config/ENVIRONMENT.yaml where ENVIRONMENT is an
+// environment variable specifing "staging" or "production", else it defaults to
+// "development".
```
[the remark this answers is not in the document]
daniel-nichter: Fixed.

request-manager/auth/auth.go
daniel-nichter: Done

request-manager/auth/auth.go
```diff
@@ -1,3 +1,9 @@
+// Copyright 2018, Square, Inc.
+
+// Package auth provides request authentication and authorization. By default,
+// there is no auth; all callers and requests are allowed. The Plugin interface
+// allows user-defined auth in combination with user-defined request sepc ACLs.
```
felixplajer: "request *spec* ACLs"

request-manager/auth/auth.go
```diff
@@ -43,12 +71,21 @@ func (a AllowAll) Authorize(c Caller, op string, req proto.Request) error {
...
+	// Role grants admin access (all ops) to request. Mutually exclusive with Ops.
+	Admin bool
+
+	// Ops granted to this role. Mutually exclusive with Admin.
+	Ops []string
 }
 
-// Manager provides request authorization using the auth plugin and a list of ACLs.
+// Manager provides pre- and post-authorization. It matches caller ACLs to request
+// ACLs, and calls Plugin.Authorize when there is a match. It also handles admin
```
felixplajer: It matches caller roles, not ACLs, to request ACLs, right?

request-manager/auth/auth.go
```diff
@@ -65,14 +102,20 @@ func NewManager(plugin Plugin, acls map[string][]ACL, adminRoles []string, stric
...
-// args. For example, the caller might not be allowed to run the request on certain
-// hosts. If authorized, returns nil; else, returns an error.
+// Authorize authorizes the request based on its ACLs, if any. If the Caller has
+// an admin role, it is allowed immediately (no further checks). Else, Caller
+// roles are matched to request ACL roles. On match, the op is matched to the
+// request ACL ops. On match again, the Plugin Authorize method is called for
+// post-authorization. If it returns nil, the request is allowed.
+//
+// If the request has no ACLs and the caller does not have an admin role, strict
+// mode determines the result: allow if disabled (no ACLs = allow allow), deny
```
felixplajer: "(no ACLs = allow *all*)"

request-manager/auth/auth.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/square/spincycle/proto"
+)
+
+// Caller represents an HTTP client making a request. Callers are determined by
+// the Plugin Authenticate method. The default Plugin (AllowAll) returns a zero
+// value Caller.
+type Caller struct {
+	// Name of the caller, whether human (username) or machine (app name). The
+	// name is user-defined and only used by Spin Cycle for logging and setting
+	// proto.Request.User.
```
felixplajer: If we're keeping the SetUsername hook for now, the caller Name isn't being used for proto.Request.User.

PR #65, opened 2018-11-30
Request Manager suspends + resumes requests

felixplajer:
- Added a request "Resumer" to the RM, which saves incoming Suspended Job Chains from the Job Runner to a table and periodically tries to resume all the SJCs it has stored. This is safe for multiple RM instances (and so multiple Resumers) running at once.
- Added graceful shutdown capability to the RM, specifically the API and request resumer.
- Added a 'jr_host' column to the requests table, which stores the specific Job Runner instance that's running a request. This'll be used to direct live status requests to the correct JR when there are mutliple JR instances running (but that's not in this PR).


job-runner/client.go
```diff
@@ -41,52 +43,64 @@ func NewClient(c *http.Client, baseUrl string) Client {
 	}
 }
 
-func (c *client) NewJobChain(jobChain proto.JobChain) error {
+func (c *client) NewJobChain(jobChain proto.JobChain) (string, error) {
```
daniel-nichter: This is a backwards-incompatible change. I.e. it'll break all code using `spincycle.Client`, which for us is a fair amount of code. We might have to +1 the minor version to control a roll out of this while we upgrade our code. Unless we can avoid making this breaking change?

proto/s2s.go
```diff
@@ -49,6 +49,10 @@ type Request struct {
 	JobChain     *JobChain `json:",omitempty"`   // the job chain
 	TotalJobs    int       `json:"totalJobs"`    // the number of jobs in the request's job chain
 	FinishedJobs int       `json:"finishedJobs"` // the number of finished jobs in the request
+
+	// The Job Runner running this request. This is a pointer so it can have a nil
+	// value when the request isn't running.
+	JobRunnerHost *string `json:"jrHost,omitempty"`
```
daniel-nichter: `*string` should be rare in Go imoh. In this case, empty string is fine, and `omitempty` will work as expected. This makes dealing with the field consistent, i.e. .it's always a string, just:
```
if req.JobRunnerHost == "" {
    // req isn't running anywhere
}
```

request-manager/request/manager.go
```diff
@@ -180,8 +183,54 @@ func (m *manager) Create(reqParams proto.CreateRequestParams) (proto.Request, er
...
+		&req.Id,
+		&req.Type,
+		&req.State,
+		&user,
+		&req.CreatedAt,
+		&startedAt,
+		&finishedAt,
+		&req.TotalJobs,
+		&req.FinishedJobs,
+		&req.JobRunnerHost); err != nil {
```
daniel-nichter: nit: break this line so this whole thing doesn't read as one

request-manager/request/manager.go
```diff
@@ -353,6 +394,29 @@ func (m *manager) Status(requestId string) (proto.RequestStatus, error) {
 	return reqStatus, nil
 }
 
+func (m *manager) Finish(requestId string, finishParams proto.FinishRequestParams) error {
+	req, err := m.Get(requestId)
+	if err != nil {
+		return err
+	}
+
+	now := time.Now()
```
daniel-nichter: Should JR set finish time? If there's a delay sending finished req back to RM, this will skew the finish time.

request-manager/request/manager.go
```diff
@@ -353,6 +394,29 @@ func (m *manager) Status(requestId string) (proto.RequestStatus, error) {
...
+	req.FinishedAt = &now
+	prevState := req.State
+	req.State = finishParams.State
+
+	// This will only update the request if the current state is RUNNING.
+	err = m.updateRequest(req, proto.STATE_RUNNING)
+	if err != nil {
+		if prevState != proto.STATE_RUNNING {
```
daniel-nichter: Would this ever happen here? Why?

request-manager/resources/request_manager_schema.sql
```diff
@@ -41,3 +42,14 @@ CREATE TABLE IF NOT EXISTS `job_log` (
 
   PRIMARY KEY (`request_id`, `job_id`, `try`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+CREATE TABLE IF NOT EXISTS `suspended_job_chains` (
+  `request_id`          BINARY(20) NOT NULL,
+  `suspended_job_chain` LONGBLOB NOT NULL,
+  `rm_host`             VARCHAR(64) NULL DEFAULT NULL,
```
daniel-nichter: Why is the Request Manager host saved?

request-manager/resources/request_manager_schema.sql
```diff
@@ -41,3 +42,14 @@ CREATE TABLE IF NOT EXISTS `job_log` (
...
+  `last_updated_at`     TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
```
daniel-nichter: Typically, this is called just `update_at`. "Last" is implied.

request-manager/server/server.go
```diff
@@ -26,6 +30,45 @@ func NewServer(appCtx app.Context) *Server {
 	}
 }
 
+// Run runs the Request Manager API in the foreground. It returns when the API stops.
+func (s *Server) Run() error {
+	if s.api == nil {
+		panic("Server.Run called before Server.Boot")
+	}
+
+	go s.waitForShutdown()
```
daniel-nichter: Wait for shutdown after starting? Else, in pathological case, we could shutdown before we've started.

request-manager/request/resumer.go
```diff
@@ -0,0 +1,606 @@
...
+
+	jr "github.com/square/spincycle/job-runner"
+	"github.com/square/spincycle/proto"
+	"github.com/square/spincycle/request-manager/db"
+)
+
+type Resumer interface {
+	// Run periodically attempts to resume any Suspended Job Chains stored in
+	// the database, and also cleans up any SJCs that get into a bad state. Run
+	// blocks until the Resumer is shut down by closing shutdownChan.
```
daniel-nichter: `shutdownChan` is an implementation detail. Better to document it at the object that implements this interface.

request-manager/request/resumer.go
```diff
@@ -0,0 +1,606 @@
...
+	q := "SELECT requests.request_id, requests.state, sjcs.suspended_job_chain FROM suspended_job_chains sjcs JOIN requests ON sjcs.request_id = requests.request_id WHERE sjcs.rm_host IS NULL ORDER BY RAND()"
+	rows, err := conn.QueryContext(ctx, q)
+	if err != nil {
+		log.Errorf("error querying db for SJCs: %s", err)
+		return
+	}
+	defer rows.Close()
+
+	var requests []proto.Request
+	sjcs := make(map[string]proto.SuspendedJobChain) // request id -> SJC
```
daniel-nichter: `sjcs := map[string]proto.SuspendedJobChain{}` is more common.

request-manager/request/resumer.go
```diff
@@ -0,0 +1,606 @@
...
+	ctx := context.TODO()
+	conn, err := r.dbc.Open(ctx)
+	if err != nil {
+		log.Errorf("could not connect to database: %s", err)
+		return
+	}
+	defer r.dbc.Close(conn)
+
+	// Retrieve Request IDs + states for all (unclaimed) SJCs, in random order.
+	q := "SELECT requests.request_id, requests.state, sjcs.suspended_job_chain FROM suspended_job_chains sjcs JOIN requests ON sjcs.request_id = requests.request_id WHERE sjcs.rm_host IS NULL ORDER BY RAND()"
```
daniel-nichter: nit: break long SQL on major parts like:
```
"SELECT ..." +
" FROM tables..." +
" JOIN x ON ... " +
" WHERE ..."
```
Just makes it easier to read.

request-manager/request/manager.go
felixplajer: I looked into it a bit - as the code is now I don't think we can ever Finish a request when its state isn't running. I want to keep the check on the error, though, because if a bug's introduced later it's easier to debug `request not in __ state (state = __)` than `cannot find record to update`.

request-manager/resources/request_manager_schema.sql
felixplajer: Setting the RM host is how the resumer claims an SJC - it sets rm_host from null to its hostname.

request-manager/server/server.go
felixplajer: Sure, I can move this to after starting the resumer. It doesn't really matter - it just closes shutdownChan, which would stop the resumer right away after it started.

request-manager/request/manager.go
felixplajer: Yeah, I guess it makes sense to have the finish time be as close as possible to the end of the last job, especially now that the JR retries the FinishRequest call several times.

request-manager/resources/request_manager_schema.sql
felixplajer: `update_at` or `updated_at`?

request-manager/request/resumer.go
```diff
@@ -0,0 +1,620 @@
...
+
+// Claim that this RM will resume the SJC corresponding to the request id.
+// Returns true if SJC is successfully claimed (this RM should resume the SJC);
+// returns false if failed to claim SJC (another RM is resuming the SJC).
+// claimSJC is used to make sure that only one RM is trying to resume a specific
+// SJC at any given time.
+func (r *resumer) claimSJC(requestId string, conn *sql.Conn) (bool, error) {
+	// Start a transaction
+	ctx := context.TODO()
+	txn, err := conn.BeginTx(ctx, nil)
```
daniel-nichter: Don't need a txn in this case. A single statement with auto-commit (enabled by default) is an implicit transaction.

request-manager/request/resumer.go
```diff
@@ -0,0 +1,620 @@
...
+
+// Un-claim the SJC corresponding to the request id - indicates an RM is no longer
+// trying to resume this SJC - so it may be claimed again later.
+// strict = true requires that this RM has a claim on the SJC, or unclaimSJC
+// will return an error. Strict should always be used except when cleaning up
+// abandoned SJCs.
+func (r *resumer) unclaimSJC(requestId string, strict bool, conn *sql.Conn) error {
+	// Start a transaction
+	ctx := context.TODO()
+	txn, err := conn.BeginTx(ctx, nil)
```
daniel-nichter: Don't need txn in this case.

request-manager/request/resumer.go
```diff
@@ -0,0 +1,620 @@
...
+	if err != nil {
+		return false, err
+	}
+	defer txn.Rollback()
+
+	// Claim SJC by setting rm_host to the host of this RM. Only claim if not
+	// alread claimed (rm_host = NULL).
+	q := "UPDATE suspended_job_chains SET rm_host = ? WHERE request_id = ? AND rm_host IS NULL"
```
daniel-nichter: + `LIMIT 1` because we want to claim only 1 SJC.

request-manager/request/resumer.go
```diff
@@ -0,0 +1,620 @@
...
+	txn, err := conn.BeginTx(context.TODO(), nil)
+	if err != nil {
+		return err
+	}
+	defer txn.Rollback()
+
+	q := "DELETE FROM suspended_job_chains WHERE request_id = ? AND rm_host = ?"
+	result, err := txn.Exec(q,
+		requestId,
+		r.host)
```
daniel-nichter: nit: these can go on one line: `txn.Exec(q, requestId, r.host")`

request-manager/request/resumer.go
```diff
@@ -0,0 +1,620 @@
...
+
+	return txn.Commit()
+}
+
+// updateRequestWithTxn updates the State and JR host as set in the given request,
+// using the provided db transaction. The request is updated only if its current
+// state in the db matches the state provided.
+func (r *resumer) updateRequestWithTxn(request proto.Request, curState byte, txn *sql.Tx) error {
+	// If JobRunnerHost is empty, we want to set the db field to NULL.
+	var jrHost sql.NullString
```
daniel-nichter: There's an easier way:
``
var jrHost interface
if request.JobRunnerHost != "" {
    jrHost = request.JobRunnerHost
}
```

File: request-manager/request/resumer.go [hunk 17]
```diff
@@ -0,0 +1,620 @@
...
+	Suspend(requestId string, sjc proto.SuspendedJobChain) error
+
+	// ResumeAll tries to resume all the SJCs currently stored in the database.
+	ResumeAll()
+
+	// Resume tries to resume a single SJC given its id and a connection to the
+	// database. The SJC must be claimed (`rm_host` field for the SJC must be set
+	// to the hostname given when creating the Resumer) before calling Resume, or
+	// it will fail.
+	Resume(id string, conn *sql.Conn) error
```
daniel-nichter: `conn *sql.Conn` is implementation detail, should be removed from interface. This would precluded a full mock (or, at least, the mock would simply ignore this arg).

request-manager/request/resumer.go
```diff
@@ -0,0 +1,620 @@
...
+
+	jr "github.com/square/spincycle/job-runner"
+	"github.com/square/spincycle/proto"
+	"github.com/square/spincycle/request-manager/db"
+)
+
+type Resumer interface {
+	// Suspend marks a running request as suspended and saves the corresponding
+	// suspended job chain.
+	Suspend(requestId string, sjc proto.SuspendedJobChain) error
```
daniel-nichter: If `proto.SuspendedJobChain` has the request ID, why the first arg?

felixplajer: no good reason 😬I'll remove it.

request-manager/server/server.go
```diff
@@ -26,6 +31,56 @@ func NewServer(appCtx app.Context) *Server {
...
+		for {
+			select {
+			case <-s.appCtx.ShutdownChan:
+				break RESUMER
+			case <-time.After(10 * time.Second):
+				s.appCtx.RR.ResumeAll()
+				s.appCtx.RR.Cleanup(1 * time.Hour)
+			}
+		}
+		close(resumerDone) // indicate the resumer loop is done
```
daniel-nichter: `defer` this. Defer for uses like this is almost always the preferred Go way.

request-manager/server/server.go
```diff
@@ -26,6 +31,56 @@ func NewServer(appCtx app.Context) *Server {
...
+	resumerDone := make(chan struct{})
+	go func() {
+		// Every 10 seconds until shutdown, resume all SJCs and clean up any
+		// SJCs in a bad state.
+	RESUMER:
+		for {
+			select {
+			case <-s.appCtx.ShutdownChan:
+				break RESUMER
+			case <-time.After(10 * time.Second):
```
daniel-nichter: You probably just want to create a `time.Ticker`.

PR #68, opened 2018-12-07
make API run + shutdown configurable

felixplajer:
Need this so we can run the API by calling server.Run(), instead of getting the api with server.API() and running it ourselves, since server.Run() handles shutdown and (in the RM) runs the Resumer.

### DIFF
--- a/job-runner/app/app.go
+++ b/job-runner/app/app.go
@@ -30,6 +30,16 @@ type Hooks struct {
 	LoadConfig  func(Context) (config.JobRunner, error)
 	Auth        func(*http.Request) (bool, error)
 	SetUsername func(*http.Request) (string, error)
+
+	// RunAPI runs the Job Runner API. It should block until the API is stopped
+	// via a call to StopAPI. If this hook is provided, it is called instead of
+	// api.Run, and StopAPI must be provided as well.
+	RunAPI func() error
+
+	// StopAPI stops running the Job Runner API. It's called when the server is
+	// stopped, and it should cause RunAPI to return. If this hook is provided, it
+	// is called instead of api.Stop, and RunAPI must be provided as well.
+	StopAPI func() error
 }
 
 func Defaults() Context {
--- a/job-runner/server/server.go
+++ b/job-runner/server/server.go
@@ -48,10 +48,10 @@ func NewServer(appCtx app.Context) *Server {
 // hook has been provided, it will be called to run the API instead of the default
 // api.Run.
 //
-// If autoShutdown = true, the server will listen for TERM and INT signals from the
-// OS and call Stop to shut itself down when those signals are received. Else, the \
+// If stopOnSignal = true, the server will listen for TERM and INT signals from the
+// OS and call Stop to shut itself down when those signals are received. Else, the
 // caller must call Stop to shut down the server.
-func (s *Server) Run(autoShutdown bool) error {
+func (s *Server) Run(stopOnSignal bool) error {
 	if s.api == nil {
 		panic("Server.Run called before Server.Boot")
 	}
@@ -59,9 +59,9 @@ func (s *Server) Run(autoShutdown bool) error {
 		return fmt.Errorf("server stopped")
 	}
 
-	// If autoShutdown = true, watch for TERM + INT signals from the OS and shut
+	// If stopOnSignal = true, watch for TERM + INT signals from the OS and shut
 	// down the Job Runner when we receive them.
-	if autoShutdown {
+	if stopOnSignal {
 		go s.waitForShutdown()
 	}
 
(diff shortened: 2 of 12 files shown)


job-runner/app/app.go
```diff
@@ -30,6 +30,16 @@ type Hooks struct {
...
+
+	// RunAPI runs the Job Runner API. It should block until the API is stopped
+	// via a call to StopAPI. If this hook is provided, it is called instead of
+	// api.Run, and StopAPI must be provided as well.
+	RunAPI func() error
+
+	// StopAPI stops running the Job Runner API. It's called when the server is
+	// stopped, and it should cause RunAPI to return. If this hook is provided, it
+	// is called instead of api.Stop, and RunAPI must be provided as well.
+	StopAPI func() error
```
daniel-nichter: Nice code comments/explanations. Thanks!

job-runner/server/server.go
```diff
@@ -21,31 +24,81 @@ import (
...
-		return err
+// Run runs the Job Runner API in the foreground. It returns when the API stops
+// running (either from an error, or after a call to Stop). If a custom RunAPI
+// hook has been provided, it will be called to run the API instead of the default
+// api.Run.
+//
+// If autoShutdown = true, the server will listen for TERM and INT signals from the
+// OS and call Stop to shut itself down when those signals are received. Else, the \
+// caller must call Stop to shut down the server.
+func (s *Server) Run(autoShutdown bool) error {
```
daniel-nichter: How about `stopOnSignal bool`? Seems to describe more what it does.

job-runner/server/server.go
```diff
@@ -55,6 +108,60 @@ func (s *Server) Boot() error {
...
+	for !s.traverserRepo.IsEmpty() {
+		select {
+		case <-time.After(10 * time.Millisecond):
+			// Check again if traversers are all done.
+		case <-timeout:
+			break WAIT_FOR_TRAVERSERS
+		}
+	}
+
+	// Stop the API, using the StopAPI hook if provided and api.Stop otherwise.
```
daniel-nichter: Do we need to stop API first? I.e. so it won't accept any new work.

request-manager/request/resumer.go
```diff
@@ -37,10 +37,10 @@ type Resumer interface {
...
 	// while, meaning the RM resuming them probably crashed. These SJCs are
 	// unclaimed (set `rm_host` to null) so they can be resumed in the future. Old
-	// SJCs are those which haven't been resumed within the TTL provided (rounded
-	// to the nearest second). They're deleted and their requests' states set to
-	// FAILED.
-	Cleanup(ttl time.Duration)
+	// SJCs are those which haven't been resumed within the TTL provided when
+	// creating the Resumer (rounded to the nearest second). They're deleted and
+	// their requests' states set to FAILED.
+	Cleanup()
```
daniel-nichter: The TTL was removed but the code comment still references it?

request-manager/server/server.go
```diff
@@ -19,71 +22,144 @@ import (
 	"github.com/square/spincycle/request-manager/status"
 )
 
+var (
+	// How often the request resumer is run.
+	ResumerCadence = 10 * time.Second
```
daniel-nichter: `ResumerInterval`  "interval" is the common term.

request-manager/server/server.go
```diff
@@ -141,20 +218,36 @@ func (s *Server) Boot() error {
...
+
+	err := s.Stop()
+	if err != nil {
+		log.Errorf("error shutting down server: %s", err)
+	}
+}
+
 // MapACL maps spec file ACL to auth.ACL structure.
-func MapACL(specs grapher.Config) map[string][]auth.ACL {
+func mapACL(specs grapher.Config) map[string][]auth.ACL {
```
daniel-nichter: Why private? I think there were unit tests for this.

job-runner/server/server.go
felixplajer: When we close the shutdownChan (line 134), the API stops accepting new job chains to run (also SJCs to resume). There's a comment there to that effect.

request-manager/request/resumer.go
felixplajer: The TTL is now provided when creating the Resumer (it's in the ResumerConfig passed to NewResumer), hence "the TTL provided when creating the Resumer".

request-manager/server/server.go
daniel-nichter: If no tests then I had intended to test it but didn’t actually do it. It’s fine for now. It needs to be refactored/moved anyway.

job-runner/server/server.go
daniel-nichter: 👍

PR #75, opened 2019-01-03
Send stop and status requests to a specific JR instance

felixplajer:
When there are multiple Job Runner instances running, the Request Manager needs to query the Job Runner running a specific request to get its real-time status or stop it. A Job Runner returns its location (a url) when starting a job chain, which the Request Manager stores and uses instead of the default JR API url when doing status and stop queries.

The new ServerURL hook in the app context defines the url a Job Runner instance returns. By default, this is just the ListenAddress specified in the config, but a custom ServerURL func can be provided if ListenAddress won't be set.

There's a small schema change in the RM db - "jr_host" to "jr_url" to better reflect what that field will be storing.

### DIFF
--- a/job-runner/api/api.go
+++ b/job-runner/api/api.go
@@ -185,7 +185,7 @@ func (api *API) newJobChainHandler(c echo.Context) error {
 	}
 
 	// Set the location in the response header to point to this server.
-	c.Response().Header().Set("Location", api.baseURL)
+	c.Response().Header().Set("Location", api.chainLocation(jc.RequestId))
 
 	// Start the traverser, and remove it from the repo when it's
 	// done running. This could take a very long time to return,
@@ -229,7 +229,7 @@ func (api *API) resumeJobChainHandler(c echo.Context) error {
 	}
 
 	// Set the location in the response header to point to this server.
-	c.Response().Header().Set("Location", api.baseURL)
+	c.Response().Header().Set("Location", api.chainLocation(sjc.RequestId))
 
 	// Start the traverser, and remove it from the repo when it's
 	// done running. This could take a very long time to return,
@@ -304,6 +304,10 @@ func (api *API) statusRunningHandler(c echo.Context) error {
 }
 
 // ------------------------------------------------------------------------- //
+
+func (api *API) chainLocation(requestId string) string {
+	return api.baseURL + API_ROOT + "job-chains/" + requestId
+}
 
 func handleError(err error) *echo.HTTPError {
 	switch err.(type) {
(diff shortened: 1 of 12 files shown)


job-runner/api/api.go
```diff
@@ -184,7 +185,7 @@ func (api *API) newJobChainHandler(c echo.Context) error {
 	}
 
 	// Set the location in the response header to point to this server.
-	c.Response().Header().Set("Location", chainLocation(jc.RequestId, os.Hostname))
+	c.Response().Header().Set("Location", api.baseURL)
```
daniel-nichter: Why this change? Location should be a URI, i.e. a full path to the returned resource (the job chain, in this case), not just the base URL.

request-manager/status/manager.go
```diff
@@ -50,10 +50,43 @@ func NewManager(dbc myconn.Connector, jrClient jr.Client) Manager {
...
-	running, err := m.jrc.SysStatRunning()
+	ctx := context.TODO()
+	conn, err := m.dbc.Open(ctx)
+	if err != nil {
+		return status, err
+	}
+	defer m.dbc.Close(conn) // don't leak conn
+
+	// Make a list of the URLs of all JR hosts currently running any requests.
+	q := "SELECT jr_url FROM requests WHERE state = ? AND jr_url IS NOT NULL"
```
daniel-nichter: Could use `SELECT DISTINCT jr_url FROM...` to avoid map, i.e. make MySQL de-dupe the list. In this simple case, and having an index on `state` col already, it's probably same  query perf for MySQL. Don't have to change it, just noting.

request-manager/status/manager.go
```diff
@@ -89,9 +115,9 @@ func (m *manager) Running(f Filter) (proto.RunningStatus, error) {
 		seen[r.RequestId] = true
 	}
 
-	q := "SELECT request_id, type, state, user, created_at, started_at, finished_at, total_jobs, finished_jobs" +
+	q = "SELECT request_id, type, state, user, created_at, started_at, finished_at, total_jobs, finished_jobs" +
 		" FROM requests WHERE request_id IN (" + db.IN(ids) + ")"
-	rows, err := conn.QueryContext(ctx, q)
+	rows, err = conn.QueryContext(ctx, q)
```
daniel-nichter: I'm not sure if/how this will work with the earlier `defer rows.Close()`.  I.e.:
```go
rows := .. // first val
defer rows.Close()
rows = ... // new val
defer rows.Close()
```
Not sure how Go handles that? Probably best to just avoid it by using `rows2` for the second call.

job-runner/api/api.go
daniel-nichter: Having reviewed rest of code, I think I see why this was changed: so RM can set column `jr_url` = to this Location. I think it's better to keep Location true to purpose (full URL of resource), and in RM use https://golang.org/pkg/net/url/#Parse to parse out the parts needed.

daniel-nichter: https://play.golang.org/p/TYR4RcTUzgV might do the trick

PR #81, opened 2019-01-28
Recover from job panics

felixplajer:
Recover from a panic in job.Run(). Panics are treated like any other job failure - the job + sequence are retried as normal.

### DIFF
--- a/job-runner/runner/runner.go
+++ b/job-runner/runner/runner.go
@@ -4,6 +4,7 @@
 package runner
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


job-runner/runner/runner.go
```diff
@@ -190,6 +188,34 @@ TRY_LOOP:
...
 
+// Actually run the job.
+func (r *runner) runJob(jobData map[string]interface{}) (startTime, finishTime int64, ret job.Return, err error) {
+	var startedAt, finishedAt int64
+
+	defer func() {
+		// Recover from a panic inside Job.Run()
+		if panicErr := recover(); panicErr != nil {
+			// Set return values.
+			startTime = startedAt
```
daniel-nichter: Why reassigning the value? Since it's set before running, it should always be set.

job-runner/runner/runner.go
```diff
@@ -190,6 +188,34 @@ TRY_LOOP:
...
+			finishTime = time.Now().UnixNano()
+			ret = job.Return{
+				State: proto.STATE_FAIL,
+				Exit:  1,
+			}
+			err = fmt.Errorf("panic from job.Run: %s", panicErr)
```
daniel-nichter: This makes it into JLE.Error? If so, probably good to make a quick code comment here to let future devs know why this line is here (it seems isolated and nothing doing anything, but it's important).

PR #89, opened 2019-02-14
Db retry and error handling

daniel-nichter:
* Retry db queries in request.Manager
* Add proto.Error
* Move user-facing errors to errors/ pkg
* Remove rm/db/ pkg

This makes transient `*sql.DB` errors largely transparent to callers. For example, if MySQL restarts, all the `*sql.DB` conn will be invalid and return an error on first attempt. But 2nd attempt will reconnect and work. So wrapping request.Manager queries in a retry block makes it work instead of reporting an opaque error like "driver: bad connection" to user.


request-manager/status/manager.go
```diff
@@ -55,7 +55,7 @@ func (m *manager) Running(f Filter) (proto.RunningStatus, error) {
 	q := "SELECT jr_url FROM requests WHERE state = ? AND jr_url IS NOT NULL"
 	rows, err := m.dbc.QueryContext(ctx, q, proto.STATE_RUNNING)
 	if err != nil {
-		return status, err
+		return status, serr.NewDbError(err, "SELECT requests")
```
michaelfinch: Why no retries on these queries?

daniel-nichter: Because lazy programmer :-)

PR #90, opened 2019-02-17
Change jekyll theme to https://pmarsceill.github.io/just-the-docs/

michaelfinch:
What it looks like:
image


daniel-nichter:
Those header links are super helpful, e.g. https://square.github.io/spincycle/basic-concepts#request-specs

michaelfinch:
What about this: https://pmarsceill.github.io/just-the-docs/docs/ui-components/buttons/#using-utilities-with-buttons? If you click on anything in the table of contents at the top of the page, you get a link to the header. Or do you want to be able to click directly on the header itself and get a link?

daniel-nichter:
The latter: want a link to a section from that section. Any TOC would have to work that way, so really this is is about linking from the other end, so to speak, because people don't always go through/start with TOC.

daniel-nichter:
Let's just merge this for now and fix it later. It's just a feature, but a really helpful one.

PR #97, opened 2019-03-07
Refactor, test, and fix spinc help

daniel-nichter:
* Fix "spinc help reqeust" (lists request args)
* Add "spinc help command"
* Make help and version real cmd like others
* Factor all help stuff out of config and into help cmd
* **Make --addr default to localhost:32308**
* Remove unused dep
* Remove --ping
* Fix --timeout (it wasn't used, now it is)


spinc/spinc.go
```diff
@@ -51,155 +46,105 @@ func Run(ctx app.Context) error {
...
+		app.Debug("%s Run error: %s", c.Cmd, err)
 	}
-
-	return nil
+	return err
 }
 
-func makeRMC(ctx *app.Context) (rm.Client, error) {
+func makeRMC(ctx app.Context) (rm.Client, error) {
 	if ctx.Options.Addr == "" {
```
felixplajer: There's a reference to --ping in the error message, and also it should be impossible for Addr to be empty now.

spinc/cmd/status.go
```diff
@@ -75,3 +77,7 @@ func (c *Status) Run() error {
 func (c *Status) Cmd() string {
 	return "status " + c.reqId
 }
+
+func (c *Status) Help() string {
+	return "'spinc status <request ID>' prints the status of the request.\n"
```
felixplajer: I'd mention -v for verbose output (i.e. running job status).

spinc/cmd/ps.go
```diff
@@ -123,3 +123,7 @@ func (c *Ps) Run() error {
 func (c *Ps) Cmd() string {
 	return "ps"
 }
+
+func (c *Ps) Help() string {
+	return "'spinc ps' prints all running requests and jobs.\n"
```
felixplajer: Would also mention `-v` here.

spinc/spinc.go
daniel-nichter: There’s a hook that can modify the options, so technically it could unset addr. But it's unclear why it would do that, so I can move the defaults after the hook to ensure it can't shoot itself in the foot.

spinc/cmd/status.go
daniel-nichter: Let's hold off on that because I'm going to rework the spinc output and remove `-v`.

PR #98, opened 2019-03-08
Refactor JR to prep for sending RM finished_jobs count

daniel-nichter:
* Remove chain Redis repo
* Factor out Chain.Validate to chain.Validate (pkg func) and call in API contorller
* Fix bug: chainRepo.Add() was never called
* Remove all calls to chainRepo.Set(): not needed with in-memory repo
* Move chainRepo.Remove() from reapers to traverser because the latter adds it
* Rename Job.Sequence.Id tag `sequenceStartId` to `sequenceId` (it's "sequenceId" in the code)
* Add `proto.JobStatus.Try`


job-runner/chain/reaper.go
```diff
@@ -210,7 +207,6 @@ func (r *RunningChainReaper) Reap(job proto.Job) {
 
 	// Set the final state of the job in the chain.
 	r.chain.SetJobState(job.Id, job.State)
-	r.chainRepo.Set(r.chain)
```
michaelfinch: What's the tldr; of removing all these chainRepo calls?

PR #99, opened 2019-03-13
Finished jobs refactoring

daniel-nichter:
⭐️ = changes for finished jobs

# Job Runner

## chain.Validate
* Check job states
* Check FinishedJobs count ⭐️

## chain.Chain
* Removed `numJobsRun` (aka `proto.JobStatus.N`). New source of truth: finished jobs count.
* Removed chain validation in `New`. API controller validates chain by calling `chain.Validate` so `Chain.New` can trust the given `proto.JobChain`.
* **`IsDoneRunning`: Make FAIL explicit case that does not check `isRunnable`. panic on invalid states. Add more comments. See func code comment for explanation.**
* **`isRunnable`: Remove job try logic for STOPPED jobs. This is done in `MakeFromSJC` now.**
* Add `FailedJobs`: stop and suspend reapers use this to determine if the chain failed (has any failed jobs), in which case it sets final chain state = FAIL rather than STOPPED or SUSPENDED, respectively.
* `AddJobTries` and `SetLatestRunJobTries` combined into `IncrementJobTries`: the new func takes a delta and, if positive, increments both total and current job tries. If negative, it only decrements current job tries because total can only increase.
* Replace `TotalTries` with `JobTries`: new func returns current and total job tries.
* `IncrementSequenceTries` now takes a `delta int` (to match other Increment* funcs).
* Add `IncrementFinishedJobs` ⭐️ 
* Remove `N` logic from `SetJobState` (being replaced by finished jobs count) ⭐️
* Remove `chainJSON`: no longer needed since Redis chain repo removed

## chain.Reaper
* Call `IncrementFinishedJobs` in all reapers when job is done with state = COMPLETE ⭐️
* Put all seq retry logic in `prepareSequenceRetry` and make it return seq start job ⭐️
* Change `Finalize`: if chain is complete, mark it COMPLETE; or if any jobs failed `chain.FaildJobs() > 0`, then mark it FAIL; else mark it STOPPED, SUSPENDED, or COMPLETE (depending on reaper)
* `sendFinalState`: send a `proto.FinishRequest` with `FinishedJobs = chain.FinishedJobs()` ⭐️ 
* `prepareSequenceRetry`: Workaround bug so that list of seq jobs to retry is unique. Count number of completed jobs in the list and decrement finished jobs by that number (i.e. keep finished jobs up to date) ⭐️

## chain.Traverser
* `runJobs` overhaul to avoid race conditions and handle sequence try count and job try count on resume
* Add `pending` and `pendingChan`: For cleaner shutdown in `stopRunningJobs`. The chan is closed when `runJobs` returns, so we know there won't be anymore goroutines for jobs launched. And `pending` is a wait group (but used sync.Atomic so it's nonblocking) that counts job routines which are going to join runner.Repo. When `pending == 0`, all job goroutines are in the runner.Repo.
* Add `stopChan`: closed early in `Stop` and `shutdown` to make `runJobs` switch from launching job goroutines to noop (and drain `runJobChan`). This fixes non-thread-safe check on `t.stopped` and `t.suspended`.
* Pass timeout to `stopRunningJobs`. 
* `stopRunningJobs`: wait for `pendingChan` then `pending == 0`
* **In `MakeFromSJC`, prep chain to be read-to-run again: set STOPPED -> PENDING and roll back try counts**

## runner
* Handle stop before running, during running, and after run during retry wait. On stop, override state = STOPPED.
* Remove `sequenceRetry uint` from `Factory.Make`
* Remove `pJob.Retry -= prevSeqTries`: instead of decreasing retry count, set try count to actual previous try count
* Remove `SequenceId` and `SequenceTry` from JLE

## proto
* Remove `STATE_TIMEOUT`, replace with `STATE_RESERVED` to keep state numbers the same. FAIL is the only not-successful state.
* Add `JobChain.FinishedJobs` ⭐️
* Change `Request.TotalJobs` and `FinishedJobs` from int to uint (they can't be negative)
* Remove `JobLog.Sequence*`
* Remove `JobStatus.N` ⭐️
* Add `RequestProgress`: JR sends to RM every 1s to update finished jobs ⭐️
* Add `FinishRequest.FinishedJobs` ⭐️

# Request Manager
* Add `requests/:reqId/progress` API endpoint ⭐️
* Remove call to `rm. IncrementFinishedJobs` in create JLE handler ⭐️
* Add `UpdateProgress` method to rm.Client for use by JR ⭐️
* Add `UpdateProgress` method to status.Manager ⭐️
* Changed `FinishRequest` method to take `proto.FinishRequest` arg ⭐️
* Remove `Sequence*` from joblog.Store

# spinc
* Remove sequence fields from "log" cmd output
* Use `Request.FinishedJobs` instead of `JobStatus.N` for "ps" output ⭐️


job-runner/chain/chain.go
```diff
@@ -145,19 +122,24 @@ func (c *Chain) IsDoneRunning() (done bool, complete bool) {
...
 			if c.canRetrySequence(job.Id) {
-				// This failed job is part of a sequence that can be retried.
 				return false, false
 			}
+			// Failed but no seq retry means the chain has failed
+		default:
+			panic("IsDoneRunning: invalid job state: " + proto.StateName[job.State])
 		}
 
 		// We can only arrive here if a job is not complete (pending or failed).
```
michaelfinch: Update this comment, we also arrive here if a job is stopped.

job-runner/chain/chain.go
```diff
@@ -92,10 +65,8 @@ func (c *Chain) NextJobs(jobId string) proto.Jobs {
 	return nextJobs
 }
 
-// IsRunnable returns whether or not a job is runnable. A job is considered
-// runnable if it is Pending or Stopped with some retry attempts remaining,
-// and all of its previous jobs are complete. If any previous jobs are not
-// complete, the job is not runnable.
+// IsRunnable returns true if the job is runnable. A job is runnable if its
+// state is PENDING or STOPPED and all immediately previous jobs are state COMPLETE.
 func (c *Chain) IsRunnable(jobId string) bool {
```
michaelfinch: Not directly related to this PR, but why do we even have the private isRunnable when all the public IsRunnable does it call isRunnable?

job-runner/chain/chain.go
```diff
@@ -145,19 +122,24 @@ func (c *Chain) IsDoneRunning() (done bool, complete bool) {
 		case proto.STATE_RUNNING:
 			// If any jobs are still running, the chain isn't done or complete.
 			return false, false
-		case proto.STATE_PENDING, proto.STATE_STOPPED:
-			// If any job can be run, the chain is not done or complete.
-			// Treat stopped jobs as pending jobs because they may be retried,
-			// as when resuming a suspended job chain.
+		case proto.STATE_STOPPED:
```
michaelfinch: I'm a little confused about this. When we suspend job chains, there could be many STOPPED jobs, correct? Then when we go to resume the job chain, won't there be a lot of STOPPED jobs in it that we rerun (i.e., there will be STOPPED jobs that are runnable)? Or do those get immediately switched to PENDING when a suspended job chain is resumed or something?

Is this scenario possible? - we suspend a job chain that results in 2 stopped jobs. Then we go to resume the job chain, which starts running those 2 jobs. The first job completes successfully before the next job has even started running, at which point if this function is called, it will return done=true, even though there's still work to be done on the current chain run.

job-runner/chain/reaper.go
```diff
@@ -547,9 +547,14 @@ type reaper struct {
 // if sending fails. It returns true if the final state was successfully sent;
 // else false.
 func (r *reaper) sendFinalState(finishedAt time.Time) {
+	fr := proto.FinishRequest{
```
michaelfinch: Ah, I remember the last bug - need to add RequestId to this struct

job-runner/chain/reaper.go
```diff
@@ -558,38 +563,71 @@ func (r *reaper) sendFinalState(finishedAt time.Time) {
...
 	// fails, then A and B are the previously completed jobs and C is the failed
 	// job. So, jobs A, B, and C will be added to sequenceJobsToRetry. D will not be
 	// added because it was never run.
-	var sequenceJobsToRetry []proto.Job
-	sequenceJobsToRetry = append(sequenceJobsToRetry, failedJob)
-	sequenceJobsCompleted := r.sequenceJobsCompleted(sequenceStartJob)
-	sequenceJobsToRetry = append(sequenceJobsToRetry, sequenceJobsCompleted...)
+	sequenceJobsToRetry := r.sequenceJobsCompleted(sequenceStartJob)
+
+	// @todo: fixme: sometimes we have failed job, sometimes we don't
```
michaelfinch: I'm not sure what needs to be fixed here

job-runner/chain/reaper.go
daniel-nichter: Thanks! Fixed.

job-runner/chain/reaper.go
daniel-nichter: Updated comment.

job-runner/chain/chain.go
daniel-nichter: First paragraph addressed in code comments for this func:
```go
// Stopped jobs are not runnable in this context (i.e. chain context). This
// function applies to the current chain run. Once a job is stopped, it cannot
// be re-ran in the current chain run. If the chain is re-ran (i.e. resumed),
// IsRunnable will return true for stopped jobs because stopped jobs are runnable
// in that context (i.e. job context).
```
In "chain context", stopped == done. But in "job context", stopped == (re)runnable. And yes, the stopped jobs are switched to state RUNNING (not PENDING).

Second paragraph: hm, yes I think that could happen. If job1 finishes and is reaped very quickly, after reaping the reaper calls chain.IsDoneRunning which will see job2 as state stopped because its state wasn't changed in runJobs yet. Solution: on SJC resume, set all stopped jobs to pending very first thing.

job-runner/chain/chain.go
daniel-nichter: `Chain.IsDoneRunning` calls `Chain.isRunnable`.  The caller locks whatever, and `isRunnable` does not lock anything.

job-runner/chain/chain.go
daniel-nichter: Done.

job-runner/chain/chain.go
daniel-nichter: Done. As we discussed in Slack: the traverser should be given a ready-to-run chain, so it stays free of any such logic. This logic is now in `MakeFromSJC` (traverser factory). Currently, we have only SuspendedJobChain which needs to be fixed up to become a normal, read-to-run JobChain, a d in future if we have other chain types, some higher-level code responsible for those chain types can fix them up before running in a traverser (else traverser would get cluttered with logic for all the different chain types).

job-runner/chain/traverser.go
```diff
@@ -383,100 +440,74 @@ func (t *traverser) runJobs() {
...
+			curTries, totalTries := t.chain.JobTries(job.Id)
+
+			runner, err := t.rf.Make(job, t.chain.RequestId(), curTries, totalTries)
 			if err != nil {
 				// Problem creating the job runner - treat job as failed.
 				// Send a JobLog to the RM so that it knows this job failed.
+				atomic.AddInt64(&t.pending, -1)
 				job.State = proto.STATE_FAIL
 				err = fmt.Errorf("problem creating job runner: %s", err)
 				t.sendJL(job, err)
```
felixplajer: This isn't a bug from you, but if creating the runner fails on a resumed or retried job (i.e. total job tries is already > 0, and we've already sent some logs for this job), then we're sending a duplicate job log (one with this Try # has already been sent).
eg: job was run once and failed, so total job tries = 1 and the RM already has a log with `Try: 1`. We retry the sequence and this time around we get an error creating runner, so we send another job log with `Try: 1`.
A simple fix seems to be incrementing job tries by 1 here.

job-runner/chain/traverser.go
```diff
@@ -534,41 +565,90 @@ func (t *traverser) shutdown() {
...
+
+	// Wait for in-flight runJob goroutines to add themselves to runner repo
+	if n := atomic.LoadInt64(&t.pending); n > 0 {
+		for atomic.LoadInt64(&t.pending) > 0 {
+			select {
+			case <-timeout:
+				return fmt.Errorf("stopRunningJobs: timeout waiting for pending count")
+			default:
+				time.Sleep(100 * time.Millisecond)
+			}
```
felixplajer: The nit-iest nit: In cases like this I generally prefer
```
select {
case <-timeout:
    return fmt.Errorf("stopRunningJobs: timeout waiting for pending count")
case <-time.After(100 * time.Millisecond):
}
```
so the timeout can still happen during the wait. It matters less when the wait is on the order of milliseconds though.

job-runner/chain/validate.go
```diff
@@ -16,7 +18,7 @@ func (e ErrInvalidChain) Error() string {
 }
 
 // Validate checks if a job chain is valid. It returns an error if it's not.
-func Validate(jobChain proto.JobChain) error {
+func Validate(jobChain proto.JobChain, new bool) error {
```
felixplajer: I'd add an explanation of "new" in the comment.

job-runner/runner/runner.go
```diff
@@ -128,22 +131,30 @@ TRY_LOOP:
...
-			State:       jobRet.State,
-			Exit:        jobRet.Exit,
-			Error:       errMsg,
-			Stdout:      jobRet.Stdout,
-			Stderr:      jobRet.Stderr,
+			RequestId:  r.reqId,
+			JobId:      r.jobId,
+			Name:       r.jobName,
+			Type:       r.jobType,
+			Try:        r.totalTries + tryNo, // job_log.try is monotonically increasing across all seq tries
```
felixplajer: Won't this result in some weirdness (skipped Try numbers) for resumed jobs? tryNo = tries before stop + tries by this runner, but totalTries already includes those tries before being stopped.

job-runner/runner/runner.go
```diff
@@ -165,27 +176,32 @@ TRY_LOOP:
...
+		// Wait between retries. Can be stopped while waiting which is why we
+		// need to increment tryNo first. At this point, we're effectively on
+		// the next try. E.g. try 1 fails, we're waiting for try 2, then we're
+		// stopped: we want try 2 state = stopped so that on resume try 2 is re-ran.
 		select {
 		case <-time.After(r.retryWait):
 		case <-r.stopChan:
-			// runner has been stopped
+			tryLogger.Infof("job stopped while waiting to run try %d", tryNo)
 			break TRY_LOOP
```
felixplajer: Hmmmmm, because we return with state Failed here (from the last run), I'm pretty sure the reaper will prepare a sequence retry, even if there are some job-level retries left. Setting final state = Stopped should fix this.

job-runner/runner/runner.go
felixplajer: I think you can do `r.totalTries + tryNo - r.tries`

job-runner/chain/chain.go
```diff
@@ -92,10 +65,8 @@ func (c *Chain) NextJobs(jobId string) proto.Jobs {
 	return nextJobs
 }
 
-// IsRunnable returns whether or not a job is runnable. A job is considered
-// runnable if it is Pending or Stopped with some retry attempts remaining,
-// and all of its previous jobs are complete. If any previous jobs are not
-// complete, the job is not runnable.
+// IsRunnable returns true if the job is runnable. A job is runnable if its
+// state is PENDING or STOPPED and all immediately previous jobs are state COMPLETE.
```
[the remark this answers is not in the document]
daniel-nichter: Done

job-runner/chain/chain.go
```diff
@@ -104,7 +75,7 @@ func (c *Chain) IsRunnable(jobId string) bool {
 
 // RunnableJobs returns a list of all jobs that are runnable. A job is
 // runnable if all of its previous jobs are complete and it is Pending
-// or Stopped with some retries still remaining.
+// or Stopped.
```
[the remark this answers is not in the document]
daniel-nichter: Done

job-runner/chain/chain.go
```diff
@@ -116,27 +87,33 @@ func (c *Chain) RunnableJobs() proto.Jobs {
...
-// IsDoneRunning returns two booleans - the first indicates whether the chain
-// is done, and the second indicates whether the chain is complete.
+// IsDoneRunning returns two booleans: done indicates if there are running or
+// runnable jobs, and complete indicates if all jobs finished successfully
+// (STATE_COMPLETE).
+//
+// A chain is complete iff every job finished successfully (STATE_COMPLETE).
+//
+// A chain is done running if there are no running or runnable jobs.
+// We wait for running jobs to reap them. Reapers roll back failed jobs if the
```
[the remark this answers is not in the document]
daniel-nichter: Done

job-runner/chain/chain.go
```diff
@@ -116,27 +87,33 @@ func (c *Chain) RunnableJobs() proto.Jobs {
...
-// A chain is done running if there are no jobs in it running and there are no more
-// jobs in it that can be run. This happens if all of the jobs in the chain are
-// complete, or if some or all of the jobs in the chain failed. Note that one
-// failed job does not mean the chain is done - there may still be pending jobs
-// independent of this failed job that can be run. Stopped jobs are treated the
-// same as pending jobs - they can be rerun (as they are when a suspended chain is
-// resumed).
+// Stopped jobs are not runnable in this context (i.e. chain context). This
+// function applies to the current chain run. Once a job is stopped, it cannot
+// be re-ran in the current chain run. If the chain is re-ran (i.e. resumed),
```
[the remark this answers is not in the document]
daniel-nichter: Hah done

job-runner/chain/chain.go
```diff
@@ -116,27 +87,33 @@ func (c *Chain) RunnableJobs() proto.Jobs {
...
+// IsRunnable will return true for stopped jobs because stopped jobs are runnable
+// in that context (i.e. job context).
 //
-// A chain is complete if every job in it completed successfully.
+// IsDoneRunning and IsRunnable are closely related but differ on stopped jobs.
```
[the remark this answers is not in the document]
daniel-nichter: Done

job-runner/chain/chain.go
```diff
@@ -145,31 +122,51 @@ func (c *Chain) IsDoneRunning() (done bool, complete bool) {
...
 				return false, false
 			}
+			// Failed but no seq retry means the chain has failed
+		default:
+			panic("IsDoneRunning: invalid job state: " + proto.StateName[job.State])
 		}
 
-		// We can only arrive here if a job is not complete (pending or failed).
+		// We can only arrive here if a job is not complete, stopped, or failed
+		// and sequence cannot be retried.
```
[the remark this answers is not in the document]
daniel-nichter: Done

job-runner/chain/chain.go
```diff
@@ -189,12 +186,35 @@ func (c *Chain) CanRetrySequence(jobId string) bool {
...
 
-func (c *Chain) IncrementSequenceTries(jobId string) {
+func (c *Chain) IncrementJobTries(jobId string, delta int) {
+	c.triesMux.Lock()
+	if delta > 0 {
+		// Total job tries can only increase. This is the job try count
+		// that's monotonically increasing across all sequence retries.
+		c.totalJobTries[jobId] += uint(delta)
+	}
+	// Job count wrt current sequnce try can reset to zero
```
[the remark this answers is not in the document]
daniel-nichter: Done

job-runner/chain/chain.go
```diff
@@ -207,29 +227,25 @@ func (c *Chain) SequenceTries(jobId string) uint {
...
-	defer c.triesMux.Unlock()
-	c.latestRunJobTries[jobId] = tries
-}
-
-func (c *Chain) TotalTries(jobId string) uint {
-	c.triesMux.RLock()
-	defer c.triesMux.RUnlock()
-	return c.totalJobTries[jobId]
+// IncrementFinishedJobs increments the finished jobs count by delta. Negative delta
+// is given on sequence retry. Returns the new finished jobs count.
```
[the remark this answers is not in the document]
daniel-nichter: Fixed

job-runner/chain/reaper.go
```diff
@@ -387,28 +375,33 @@ func (r *SuspendedChainReaper) Reap(job proto.Job) {
...
 	finishedAt := time.Now().UTC()
 
+	log.Infof("SuspendedChainReaper.Finalize: call")
+	defer log.Infof("SuspendedChainReaper.Finalize: return")
+
 	// Mark any jobs that didn't respond to Stop in time as Failed
-	for _, jobStatus := range r.chain.Running() {
+	for jobId, jobStatus := range r.chain.Running() {
+		r.logger.Infof("job %s still running, setting state to FAIL", jobId)
 		jobId := jobStatus.JobId
```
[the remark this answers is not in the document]
daniel-nichter: Done

job-runner/chain/reaper.go
```diff
@@ -387,28 +375,33 @@ func (r *SuspendedChainReaper) Reap(job proto.Job) {
...
+	if complete {
+		r.logger.Infof("job chain complete")
+		r.chain.SetState(proto.STATE_COMPLETE)
 		r.sendFinalState(finishedAt)
 		return
 	}
 
-	// Chain isn't done - send SuspendedJobChain to RM.
-	r.logger.Infof("chain is not done - sending Suspended Job Chain to RM...")
+	if r.chain.FailedJobs() > 0 {
```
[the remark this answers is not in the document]
daniel-nichter: We didn't really decide one way or the other. Let me think more on it...

job-runner/chain/reaper.go
```diff
@@ -558,38 +564,73 @@ func (r *reaper) sendFinalState(finishedAt time.Time) {
...
+		cur, _ := r.chain.JobTries(job.Id)             // job try count for current seq
+		r.chain.IncrementJobTries(job.Id, -1*int(cur)) // decr job current tries by ^
+
+		// Roll back job state to pending so it's runnable again
 		r.chain.SetJobState(job.Id, proto.STATE_PENDING)
 	}
+
+	// Roll back finished job count. The -1 accounts for failedJob which did
+	// not incr finished jobs.
+	r.chain.IncrementFinishedJobs(-1 * finishedJobs)
```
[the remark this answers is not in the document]
daniel-nichter: Yeah, comment was out of date. Fixed.

proto/proto.go
```diff
@@ -76,6 +78,7 @@ type JobChain struct {
 	Jobs          map[string]Job      `json:"jobs"`          // Job.Id => job
 	AdjacencyList map[string][]string `json:"adjacencyList"` // Job.Id => next []Job.Id
 	State         byte                `json:"state"`         // STATE_* const
+	FinishedJobs  uint                `json:"finishedJobs"`
```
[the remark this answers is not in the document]
daniel-nichter: Done

job-runner/chain/traverser.go
```diff
@@ -94,6 +96,33 @@ func (f *traverserFactory) Make(jobChain *proto.JobChain) (Traverser, error) {
 func (f *traverserFactory) MakeFromSJC(sjc *proto.SuspendedJobChain) (Traverser, error) {
 	// Convert/wrap chain from proto to Go object.
 	chain := NewChain(sjc.JobChain, sjc.SequenceTries, sjc.TotalJobTries, sjc.LatestRunJobTries)
+	logger := log.WithFields(log.Fields{"request_id": sjc.RequestId})
+	logger.Infof("resuming request")
+
+	// Change all STOPPED jobs to PENDING. Traverser expects a ready-to-run chain.
+	// We used to change stopped -> running runJobs, but if two jobs are stopped
```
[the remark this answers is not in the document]
daniel-nichter: Done

job-runner/chain/traverser.go
```diff
@@ -200,36 +234,30 @@ func NewTraverser(cfg TraverserConfig) *traverser {
...
-		t.logger.Infof("sending initial job (%s) to runJobChan", job.Id)
-		if t.chain.IsSequenceStartJob(job.Id) {
-			// Starting a sequence, so increment sequence try count.
-			t.chain.IncrementSequenceTries(job.Id)
-			seqLogger := t.logger.WithFields(log.Fields{"sequence_id": job.SequenceId, "sequence_try": t.chain.SequenceTries(job.Id)})
-			seqLogger.Info("starting try of sequence")
-		}
+	// Enqueue all the first runnable jobs. For a new job chain (not suspended),
+	// this'll be the first job in the chain. For a resumed job chain, it'll be
+	// stopped jobs (see Chain.isRunnable).
```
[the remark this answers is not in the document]
daniel-nichter: Fixed

job-runner/chain/traverser.go
```diff
@@ -356,11 +385,39 @@ func (t *traverser) Status() (proto.JobChainStatus, error) {
...
+		// unbuffered). As long as we do not add job to runner repo, or do anything
+		// to the job, it's like the job never ran; it stays pending and tries=0.
+		//
+		// Must check before running goroutine because Run() closes runJobChan
+		// when the runningReaper is done. Then this loop will end and close
+		// pendingChan which stopRunningJobs blocks on. Since this check happens
+		// in loop not goroutine, a closed pendingChan means it's been checked
+		// for all jobs and either the job did not run or it did with pending+1
+		// because the loop want finish until running all code before the goroutine
+		// is launched.
```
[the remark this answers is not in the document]
daniel-nichter: Fixed

job-runner/chain/traverser.go
daniel-nichter: Yeah, or maybe not sending a JLE? Technically, a JLE = the job ran (successful or not), but in this can't we can't really say it ran because we didn't get that far (can't run without a runner).

job-runner/chain/traverser.go
daniel-nichter: Normally yes but with a 100ms sleep, we won't wait know any longer than the timeout as perceived by a human.

job-runner/chain/validate.go
daniel-nichter: Updated.

job-runner/runner/runner.go
daniel-nichter: Yes, fixed by using `totalTries` instead, and incrementing it in the loop, too. Will write a test case, too.

job-runner/runner/runner.go
daniel-nichter: Yes, thanks. Supposed to be a `finalState = proto.STATE_STOPPED` there.

job-runner/chain/traverser.go
felixplajer: Yeah, but if we don't send a JLE it's very confusing if someone looks back at the logs and sees "job A completed but then job B was never run?!". You could end up with a request that's Failed, but no job logs that show Failed.

PR #100, opened 2019-03-18
Begin documenting the API

michaelfinch:
Please let me know what you think of this format. Want to get feedback on a single endpoint before I go doing all of the rest.

image
image


docs/v1.0/api/api-endpoints.md
```diff
@@ -6,3 +6,77 @@ nav_order: 1
...
 # API Endpoints
+{: .no_toc }
+
+* TOC
+{:toc}
+
+---
+
+## Requests
+All things related to Spin Cycle [requests](http://localhost:4000/spincycle/v1.0/learn-more/basic-concepts#request).
```
felixplajer: Is `localhost` right?

michaelfinch: Hah, it is not. Will update

PR #101, opened 2019-03-27
spinc v2

daniel-nichter:
* **Remove chain.running.** Runner repo is single source of trust wrt what's running. And traverser repo is source of truth wrt what requests/chains are running.
* Add spinc info cmd.
* Change ps and status cmds.
* Remove proto.RequestStatus, proto.JobChainStatus, proto.JobStatuses (use Request or RunningStatus).
* Rename and refactor SysStatRunning as Running everywhere.
* Rename/refactor chain.Traverser.Status() as .Running().
* Make Runner.Status return new runner.Status struct.
* Add proto.StatusFilter so 'spinc ps reqId' fetches only that request's status.
* Fetch job status from JRs in parallel.


job-runner/chain/traverser.go
```diff
@@ -349,35 +349,25 @@ func (t *traverser) Stop() error {
...
-	for jobId, jobStatus := range runningJobs {
-		runner := activeRunners[jobId]
-		if runner == nil {
-			// The job finished between the call to chain.Running() and now,
-			// so it's runner no longer exists in the runner.Repo.
-			jobStatus.Status = "(finished)"
-		} else {
-			jobStatus.Status = runner.Status()
+func (t *traverser) Running() []proto.JobStatus {
+	runners := t.runnerRepo.Items()                       // map[string]interface{} keyed on jobId
```
felixplajer: Items() returns map[string]Runner, not map[string]interface{}

job-runner/runner/runner.go
```diff
@@ -28,6 +28,14 @@ type Return struct {
 	Tries      uint // Number of tries this run, not including any previous tries
 }
 
+type Status struct {
+	Job       proto.Job
+	StartedAt time.Time
```
felixplajer: would be good to add a comment that StartedAt here refers to when the entire Runner started, not necessarily this try of the job.

job-runner/runner/runner.go
```diff
@@ -192,13 +200,19 @@ TRY_LOOP:
...
 		tries++
 		tryNo++
-		totalTries++
+		r.totalTries++
+		r.retrySleep = true
+		r.Unlock()
 		select {
 		case <-time.After(r.retryWait):
+			r.Lock()
+			r.retrySleep = true
```
felixplajer: `r.retrySleep = false` :wink:

job-runner/runner/runner.go
```diff
@@ -269,7 +283,24 @@ func (r *runner) Runtime() float64 {
...
+func (r *runner) Status() Status {
+	// Get real-time status before locking in case it's slow
+	status := r.realJob.Status()
+
+	r.Lock()
+	defer r.Unlock()
+
+	// Indicate in real-time status if job is sleep, i.e. not truly running
+	sleeping := r.retrySleep
+	if sleeping {
```
felixplajer: What's the reason behind not just checking `if r.retrySleep`?

proto/proto.go
```diff
@@ -163,40 +164,64 @@ type JobLog struct {
...
+func (js JobStatusByStartTime) Swap(i, j int)      { js[i], js[j] = js[j], js[i] }
 
 // RequestProgress updates request progress from the Job Runner.
 type RequestProgress struct {
 	RequestId    string `json:"requestId"`
 	FinishedJobs uint   `json:"finishedJobs"` // number of jobs that ran and finished with state = STATE_COMPLETE
 }
 
+// RunningStatus represents running jobs and their requests. It is returned by
+// GET /api/v1/status/running.
```
felixplajer: Specify that it's returned by the Request Manager from /api/v1/status/running, since the JR has the same endpoint.

request-manager/status/manager.go
```diff
@@ -56,82 +42,89 @@ func NewManager(dbc *sql.DB, jrClient jr.Client) Manager {
...
+	if len(all.Jobs) == 0 {
+		return all, nil
 	}
 
-	status.Jobs = running
-
-	if len(running) == 0 {
-		return status, nil
+	switch f.OrderBy {
+	case "starttime":
```
felixplajer: proto.go says `startTime`. Might be a good idea to just do strings.ToLower().

spinc/cmd/help.go
```diff
@@ -85,13 +85,14 @@ func (c *Help) Usage() {
 		"  --help     Print help\n"+
 		"  --timeout  API timeout, milliseconds (default: %d)\n"+
 		"  --version  Print version\n"+
```
felixplajer: Would be nice to still have an empty line between Flags + Commands (aka add `\n` to the end of this line).

spinc/cmd/ps.go
```diff
@@ -48,84 +62,54 @@ func (c *Ps) Run() error {
...
+	if c.reqId != "" {
+		return "ps " + c.reqId
+	}
 	return "ps"
 }
 
 func (c *Ps) Help() string {
-	return "'spinc ps' prints all running requests and jobs.\n"
+	return "'spinc ps [request ID]' prints running requests and jobs.\n" +
+		"Request ID is optinoal. If given, only its running jobs are printed; else, all requests' running jobs are printed.\n" +
```
felixplajer: `optinoal`

job-runner/chain/traverser.go
daniel-nichter: Thanks. Fixed.

job-runner/runner/runner.go
daniel-nichter: Done.

job-runner/runner/runner.go
daniel-nichter: Hah thanks. This is why we do code reviews. :-)

job-runner/runner/runner.go
daniel-nichter: Copy-paste error. Fixed.

proto/proto.go
daniel-nichter: Done.

request-manager/status/manager.go
daniel-nichter: Done. It's done higher up, but probably best to do it where it's actually used.

spinc/cmd/help.go
daniel-nichter: There's one: it's on next line: `"\nRun spinc (no command) to lists requests\n"`

spinc/cmd/ps.go
daniel-nichter: embarrasing

PR #106, opened 2019-04-03
Add a favicon

felixplajer:
tbh I don't know exactly _why_ adding this in `_includes/head.html` makes it be the `<head>` for every page, but ¯\\\_(ツ)\_/¯ it works


michaelfinch:
A necessary hack for now. Maybe we can make a PR to just-the-docs to support specifying a favicon in  _config.yaml

PR #107, opened 2019-04-23
make job exit codes unsigned integers

felixplajer:
exit codes are stored in the job_logs table as TINYINT UNSIGNED, so they should be unsigned integers in job.Return.Exit and proto.JobLog.Exit.


daniel-nichter:
Unix exit code is an `int`, so we need to changed the SQL schema instead. Also, this code would be a backward-incompatible change, so we'd have to release v1.1 to do it.

felixplajer:
Ah, I didn't realize exit codes are actually ints / can actually be negative.

PR #120, opened 2019-08-14
Use lower-case import path for logrus

bwester:
The logrus package uses a lower-case "sirupsen" as its canonical import
path. This commit switches to that path instead of the older upper-case
"Sirupsen". No code is changed.


bwester:
This PR shouldn't affect the CI tests in any way. If a green build is required to merge this PR, I'll rebase it after #123 merges, which should fix the failures.

PR #127, opened 2020-01-24
Add an API endpoint to search previous requests

felixplajer:
Additionally:
- Convert from dep to mod (this is why the diff is huge).
- Update APIs to Echo v4.
- Increase timestamp precision from seconds to microseconds. This makes it less likely we need to fall back on request ID when ordering requests chronologically.
- Start tracking schema changes to improve update process.


request-manager/request/manager.go
```diff
@@ -493,8 +498,102 @@ func (m *manager) GetWithJC(requestId string) (proto.Request, error) {
...
+		}
+		requests = append(requests, req)
+	}
+
+	return requests, nil
+}
+
 // ------------------------------------------------------------------------- //
 
+// Return request IDs matching the filter, ordered by create time + request ID.
```
michaelfinch: Why not get the full request here so that we don't have to run two queries for each one?

request-manager/client.go
```diff
@@ -22,6 +22,10 @@ type Client interface {
 	// GetRequest takes a request id and returns the corresponding request.
 	GetRequest(string) (proto.Request, error)
 
+	// FindRequests takes a request filter and returns a list of requests
+	// matching the filter conditions, in reverse chronological order.
```
daniel-nichter: Reverse chronological = ascending or descending by ts?

felixplajer: Descending. I'll reword this a bit to be extra clear.

request-manager/request/manager.go
felixplajer: No great reason, I was just being a bit lazy.

request-manager/client.go
felixplajer: Argh, I was sorting them the wrong way round in the code and managed to write the tests wrong as well 🤦‍♂

request-manager/resources/migrations/v003_add_id_to_timestamp_indexes.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE `requests`
+  DROP INDEX `created_at`,
+  ADD INDEX (`created_at`, `request_id`);
```
michaelfinch: I think we should also drop the `state` index and replace it with (`state`, `created_at`, `request_id`). That will get used in the list view UI which shows recent requests grouped by state.

felixplajer: You're right, we'll probably be filtering by state a lot.

felixplajer: Actually, there's also no point in dropping and recreating the `created_at` index — `request_id` is the primary key so it's already in the index.

go.mod
```diff
@@ -0,0 +1,20 @@
+module github.com/square/spincycle
+
+go 1.13
+
+require (
+	github.com/alexflint/go-arg v1.0.0
+	github.com/go-sql-driver/mysql v1.4.1
```
daniel-nichter: While you're at it, might as well update this to v1.5.0.

proto/proto.go
```diff
@@ -252,6 +254,53 @@ func (j Jobs) Swap(i, k int) {
 	j[i], j[k] = j[k], j[i]
 }
 
+// RequestFilter represents optional filters when listing requests.
+type RequestFilter struct {
+	Type      string // Type of requests to return.
+	States    []byte // Request states to include.
+	Requestor string // User who made the request.
```
daniel-nichter: The return data is `user` (and column name in db the same). Any reason it's `Requestor` here? (Also, code comment says "User who...".

request-manager/api/api.go
```diff
@@ -430,17 +498,13 @@ func handleError(err error, c echo.Context) error {
 		HTTPStatus: http.StatusInternalServerError,
 	}
 
-	switch err.(type) {
-	case serr.RequestNotFound, serr.JobNotFound:
```
daniel-nichter: Go generally prefers `case` over `if-else` blocks. Any reason for this change?

Issue #115, opened 2019-06-06
Limiting the number of parallel sequences that run at once


michaelfinch:
It would be nice if parallel sequences could be limited to only run N at a time. It would look something like this in the yaml spec
```
job1:
  category: sequence
  type: the/sequence
  each:
    - foo:bar
  parallel: 1
```
Even if job1 expanded into 5 parallel sequences, only 1 would run at a time.

PR #135, opened 2020-06-25
Limit number of parallel sequences that can run at once

laureneliu:
Issue #115.


request-manager/grapher/grapher.go
```diff
@@ -342,8 +342,29 @@ func (o *Grapher) buildComponent(name string, nodeDefs map[string]*NodeSpec, nod
 				}
 
 				// Insert all components between the start and end vertices.
+				// Place in parallel if possible, otherwise serialize.
```
felixplajer, line 345: instead of saying `if possible`, explain how they're placed in parallel / serialized according to the `Parallel` field on the node (if one exists)

request-manager/grapher/grapher.go
```diff
@@ -342,8 +342,29 @@ func (o *Grapher) buildComponent(name string, nodeDefs map[string]*NodeSpec, nod
...
+				if err != nil {
+					return nil, err
+				}
+				prev := g.First
+				var count uint = 0
 				for _, c := range componentsForThisNode {
-					g.insertComponentBetween(c, g.First, g.Last)
+					currG.insertComponentBetween(c, currG.First, currG.Last)
+					count++
+					if n.Parallel != nil && count == *n.Parallel {
```
felixplajer, line 356: Rather than checking in every iteration if parallel != nil, how about doing that check just once outside the loop and storing the value in another variable which you compare to `count` here. You could handle the case where it _is_ nil by setting `parallel = len(componentsForThisNode)`.

request-manager/grapher/grapher_test.go
```diff
@@ -549,6 +561,144 @@ func TestFailCreateBadIfConditionalGraph(t *testing.T) {
 	}
 }
 
+func TestLimitParallel(t *testing.T) {
+	tf := &testFactory{}
+	sequencesFile := "../test/specs/decomm-limit-parallel.yaml"
+	cfg, _ := ReadConfig(sequencesFile)
+	omg := NewGrapher(req, tf, cfg, id.NewGenerator(4, 100))
```
felixplajer, line 568: looks like the other tests all use [funcs like this ](https://github.com/square/spincycle/blob/c30fa50b0f7ce5804aedb1bb02b10d662348b8a8/request-manager/grapher/grapher_test.go#L81-L93)to make their test graphers. For consistency's sake, can you do the same thing here?

request-manager/grapher/spec.go
```diff
@@ -93,6 +94,9 @@ func ReadConfig(configFile string) (Config, error) {
 		sequence.Name = sequenceName
 		for nodeName, node := range sequence.Nodes {
 			node.Name = nodeName
+			if node.Parallel != nil && *node.Parallel == 0 {
+				return cfg, fmt.Errorf("Parallel: 0 in sequence %s, node %s", sequence.Name, node.Name)
```
felixplajer, line 98: The best error messages tell you how to fix the problem they're reporting. So in this case, I would add `Parallel must be > 0` or something similar at the end of the message.

felixplajer: ah also, it's good to think about what your error will look like if a higher-level caller wraps it, eg `got error from X: <insert your error here>`. To that end it's usually a good idea to start with a lowercase letter (unless you're referring to a proper noun that's always capital).

request-manager/test/specs/spec-bad-parallel.yaml
```diff
@@ -0,0 +1,147 @@
...
+      decommission-instances:
+        category: sequence
+        type: decommission-instance
+        each:
+          - instances:instance # repeat for each instance in instances
+        args:
+          - expected: instances
+            given: instances
+        deps: [prep-1]
+        parallel: 0
```
felixplajer, line 60: put a comment in the yaml here to point out this is what makes the spec "bad"

request-manager/grapher/grapher.go
```diff
@@ -342,18 +342,28 @@ func (o *Grapher) buildComponent(name string, nodeDefs map[string]*NodeSpec, nod
 				}
 
 				// Insert all components between the start and end vertices.
-				// Place in parallel if possible, otherwise serialize.
-				// Each parallel component is wrapped between dummy nodes.
+				// Place at most `parallel` components per parallel supercomponent..
+				// Serialize parallel supercomponents if `parallel` exceeds number of
+				// components.
```
felixplajer, line 347: if number of components exceeds parallel, not the other way around

request-manager/grapher/grapher.go
```diff
@@ -342,18 +342,28 @@ func (o *Grapher) buildComponent(name string, nodeDefs map[string]*NodeSpec, nod
 				}
 
 				// Insert all components between the start and end vertices.
-				// Place in parallel if possible, otherwise serialize.
-				// Each parallel component is wrapped between dummy nodes.
+				// Place at most `parallel` components per parallel supercomponent..
```
felixplajer, line 345: nit: extra . here

Issue #118, opened 2019-06-18
Grapher should error if static arg has no default


felixplajer:
Recently spent a lot of time debugging a request, only to realize I accidentally put
```
static:
  - name: argName
    name: argValue
```
instead of
```
static:
  - name: argName
    default: argValue
````
The arg was given an empty string as its value - really the grapher should error if a static arg has no given value.

Side note: the YAML parsing is kind of weird here - the second `name: argValue` overwrites the `name: argName`, which seems like maybe it should be an error but I guess not ¯\\\_(ツ)\_/¯

PR #136, opened 2020-06-26
Error on empty static default arg in specs

laureneliu:
Issue #118.

### DIFF
--- a/request-manager/grapher/spec.go
+++ b/request-manager/grapher/spec.go
@@ -92,6 +92,12 @@ func ReadConfig(configFile string) (Config, error) {
 
 	for sequenceName, sequence := range cfg.Sequences {
 		sequence.Name = sequenceName
+		for _, arg := range sequence.Args.Static {
+			if arg.Default == "" {
+				return cfg, fmt.Errorf("static arg in sequence %s was not given a default", sequence.Name)
+			}
+		}
+
 		for nodeName, node := range sequence.Nodes {
 			node.Name = nodeName
 			if node.Parallel != nil && *node.Parallel == 0 {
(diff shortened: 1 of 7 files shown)


request-manager/grapher/spec.go
```diff
@@ -91,6 +91,12 @@ func ReadConfig(configFile string) (Config, error) {
 
 	for sequenceName, sequence := range cfg.Sequences {
 		sequence.Name = sequenceName
+		for _, arg := range sequence.Args.Static {
+			if arg.Default == "" {
+				return cfg, fmt.Errorf("static arg in sequence %s was not given a default", sequence.Name)
```
felixplajer, line 96: can you include the name of the arg in the error message?

request-manager/grapher/spec.go
```diff
@@ -91,6 +91,12 @@ func ReadConfig(configFile string) (Config, error) {
 
 	for sequenceName, sequence := range cfg.Sequences {
 		sequence.Name = sequenceName
+		for _, arg := range sequence.Args.Static {
```
felixplajer, line 94: Actually, could you do the same thing for Optional args? They should also always have a default.

request-manager/grapher/spec.go
```diff
@@ -91,6 +91,12 @@ func ReadConfig(configFile string) (Config, error) {
...
+			if arg.Default == "" {
```
felixplajer, line 95: Oh wait. What if someone writes `default: ""` — that's legit. Might need to do the same pointer trick as for Parallel to differentiate between a default not being provided, and an empty string. Unfortunately that means some refactoring of whatever else uses the ArgSpec type, to expect a pointer to a string rather than just a string.

Issue #122, opened 2019-08-15
set requests that couldn't be started to failed


felixplajer:
see https://github.com/square/spincycle/blob/cbda5089abd89b68e2d6879a234b17ed9a912b8f/request-manager/api/api.go#L210

If we fail to start a request (probably because we couldn't send it to the JR), set the request to Failed instead of leaving it as Pending forever.

PR #139, opened 2020-06-29
Mark failed requests as failed

laureneliu:
Issue #122.

### DIFF
--- a/request-manager/api/api.go
+++ b/request-manager/api/api.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/labstack/echo/v4"
 	"github.com/labstack/echo/v4/middleware"
+	log "github.com/sirupsen/logrus"
 
 	serr "github.com/square/spincycle/errors"
 	"github.com/square/spincycle/proto"
@@ -204,7 +205,9 @@ func (api *API) createRequestHandler(c echo.Context) error {
 	// Run (non-blocking)
 
 	if err := api.rm.Start(req.Id); err != nil {
-		api.rm.Fail(req.Id)
+		if err := api.rm.FailPending(req.Id); err != nil {
+			log.Errorf("error starting request in RM: %s", err)
+		}
 		return handleError(err, c)
 	}
 
--- a/request-manager/api/api.go
+++ b/request-manager/api/api.go
@@ -203,11 +203,8 @@ func (api *API) createRequestHandler(c echo.Context) error {
 	// ----------------------------------------------------------------------
 	// Run (non-blocking)
 
-	// TODO(felixp): if creating the request succeeded but starting it failed,
-	// mark the request as Failed. There's currently no way for a
-	// user to Start a request that's already been created, so otherwise
-	// the request will be Pending forever.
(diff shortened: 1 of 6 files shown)


request-manager/api/api.go
```diff
@@ -203,11 +203,8 @@ func (api *API) createRequestHandler(c echo.Context) error {
 	// ----------------------------------------------------------------------
 	// Run (non-blocking)
 
-	// TODO(felixp): if creating the request succeeded but starting it failed,
-	// mark the request as Failed. There's currently no way for a
-	// user to Start a request that's already been created, so otherwise
-	// the request will be Pending forever.
 	if err := api.rm.Start(req.Id); err != nil {
+		api.rm.Fail(req.Id)
```
felixplajer, line 207: Should check the error returned by Fail here (as a general rule always check the returned error). Since the API is already in the middle of handling another error, I think if an error is returned you should just log it (i.e. save it in the app's logs but don't return the error from the API).

...And of course there's no logging package in the API. So, let's use the same package we use elsewhere in spincycle — import `log "github.com/sirupsen/logrus"` at the top of the file. Then you can call `log.Errorf(...)` with args just like `fmt.Printf(...)` to log the error. For the record, what that does is very similar to just `fmt.Printf`, but most importantly it'll add a timestamp to the output.

request-manager/request/manager.go
```diff
@@ -56,6 +56,9 @@ type Manager interface {
 	// state from the proto.FinishRequest argument.
 	Finish(requestId string, finishParams proto.FinishRequest) error
 
+	// Fail a pending request (if it can't be started for some reason).
+	Fail(requestId string) error
```
felixplajer, line 60: Hmm, since this only works on requests that are pending, how about calling this `FailPending` to make that extra clear?

request-manager/request/manager.go
```diff
@@ -395,6 +398,31 @@ func (m *manager) Finish(requestId string, finishParams proto.FinishRequest) err
...
+
+	req.State = proto.STATE_FAIL
+	finishedAt := time.Now().UTC()
+	req.FinishedAt = &finishedAt
+	req.JobRunnerURL = ""
+
+	err = m.updateRequest(req, proto.STATE_PENDING)
+	if err != nil {
+		if prevState != proto.STATE_PENDING {
+			// This should never happen - we never finish a request that isn't pending.
```
felixplajer, line 417: `finish` -> `fail`

request-manager/api/api.go
```diff
@@ -204,7 +205,9 @@ func (api *API) createRequestHandler(c echo.Context) error {
 	// Run (non-blocking)
 
 	if err := api.rm.Start(req.Id); err != nil {
-		api.rm.Fail(req.Id)
+		if err := api.rm.FailPending(req.Id); err != nil {
+			log.Errorf("error starting request in RM: %s", err)
```
felixplajer, line 209: include the request id in the error message too please — will make it easier to see what exactly happened.

Issue #111, opened 2019-05-24
sequence retry wait doesn't exist


felixplajer:
In the docs, there's a retryWait option given for sequences - this doesn't actually exist. We should remove it from the docs (though it's a good idea for a future feature).

smanurung:
I think it does exist (https://github.com/square/spincycle/blob/master/job-runner/runner/runner.go#L72).

felixplajer:
That retry wait is for an individual job, not a whole sequence. A Runner acts on an individual job — sequence retry waits would have to be implemented higher up, in the Traverser or Reaper.

PR #142, opened 2020-07-01
Implement sequence retry wait

laureneliu:
Issue #111.

### DIFF
--- a/job-runner/chain/traverser.go
+++ b/job-runner/chain/traverser.go
@@ -411,6 +411,24 @@ func (t *traverser) runJobs() {
 		go func(job proto.Job) {
 			jLogger := t.logger.WithFields(log.Fields{"job_id": job.Id, "sequence_id": job.SequenceId, "sequence_try": t.chain.SequenceTries(job.Id)})
 
+			// If this is sequence start job (which currently means sequenceId == job.Id),
+			// wait for duration of SequenceRetryWait, then increment sequence try count.
+			if t.chain.IsSequenceStartJob(job.Id) {
+				if t.chain.SequenceTries(job.Id) != 0 {
+					jLogger.Infof(fmt.Sprintf("waiting %s before retrying sequence_id %s", job.SequenceRetryWait, job.SequenceId))
+					retryWait, _ := time.ParseDuration(job.SequenceRetryWait) // checked that this parses in RM
+					select {
+					case <-time.After(retryWait): // wait before retry
+					case <-t.stopChan:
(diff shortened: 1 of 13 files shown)


job-runner/chain/traverser.go
```diff
@@ -431,6 +431,10 @@ func (t *traverser) runJobs() {
 			// Increment sequence try count if this is sequence start job, which
 			// currently means sequenceId == job.Id.
 			if t.chain.IsSequenceStartJob(job.Id) {
+				if t.chain.SequenceTries(job.Id) != 0 {
+					retryWait, _ := time.ParseDuration(job.SequenceRetryWait) // checked that this parses in RM
+					time.Sleep(retryWait)
```
felixplajer, line 436: Ah, if the traverser gets stopped or shutdown (Stop() / shutdown()) while we're doing a sequence retry wait, it'll need to wait for this sleep to finish before it can finish actually stopping the traverser (and it times out after 10 seconds or something so that's not great). Unfortunately the whole stopping / suspending thing is really complicated. Anyways, I think the fix for this is:
1. move this `if` block to _before_ the defer at line 418.
2. replace time.Sleep with something like
```
select {
    case <-time.After(retryWait):
        # do nothing, we waited successfully
    case <-t.stopChan:
        jLogger.Infof("traverser was stopped - exiting sequence retry wait early and not running job")
        // before returning, decrement pending to signal to stopRunningJobs that there's
        // one less goroutine it nees to wait for.
        atomic.AddInt64(&t.pending, -1)
        return
}
```

job-runner/chain/traverser.go
```diff
@@ -411,6 +411,24 @@ func (t *traverser) runJobs() {
 		go func(job proto.Job) {
 			jLogger := t.logger.WithFields(log.Fields{"job_id": job.Id, "sequence_id": job.SequenceId, "sequence_try": t.chain.SequenceTries(job.Id)})
 
+			// If this is sequence start job (which currently means sequenceId == job.Id),
+			// wait for duration of SequenceRetryWait, then increment sequence try count.
+			if t.chain.IsSequenceStartJob(job.Id) {
+				if t.chain.SequenceTries(job.Id) != 0 {
+					jLogger.Infof(fmt.Sprintf("waiting %s before retrying sequence_id %s", job.SequenceRetryWait, job.SequenceId))
```
felixplajer, line 418: all log lines printed by `jLogger` actually already include the sequence id (look above where jLogger is created — it's given a bunch of fields), so you dont' need to include the id in the message

Issue #116, opened 2019-06-06
request specs: "set as" field for job args


felixplajer:
In request specs we have the
```
expected: argName
   given: otherArgName
```
 pattern to when passing in args to a job/sequence node. It would be extremely useful to have a similar 
```
sets: argName
  as: otherArgName
```
when specifying what args a job sets. This allows running the same type of job multiple times within a request, without one job overwriting the args another has set.

PR #143, opened 2020-07-05
Sets/arg/as

laureneliu:
Issue #116.


docs/release-notes/index.md
```diff
@@ -12,3 +12,26 @@ permalink: /release-notes
...
+### v2.0.0 (released tbd)
+
+* Request specs use backwards-incompatible sets-arg-as syntax. Script to upgrade [here](https://github.com/square/spincycle/util/reformat-yaml/).
+
+  Old syntax:
+  ```
+  node-name:
+    category: {job | sequence | conditional}
+    type: node-type
+    sets: [new-arg]
```
felixplajer, line 27: Maybe add a line with `...` between type and sets? It kind of looks like category, type, and sets are the only fields you can set for a node.

docs/release-notes/index.md
```diff
@@ -12,3 +12,26 @@ permalink: /release-notes
 ### v1.0.0 (released 2019-04-09)
 
 * First GA, production-ready release.
+
+## v2.0
+
+### v2.0.0 (released tbd)
+
+* Request specs use backwards-incompatible sets-arg-as syntax. Script to upgrade [here](https://github.com/square/spincycle/util/reformat-yaml/).
```
felixplajer, line 20: You should also describe the point of the change, i.e. that you can now rename arguments set by a node.

docs/v2.0/develop/requests.md
```diff
@@ -94,13 +97,15 @@ All node specs begin with a node name: "expand-cluster", in this case. Node name
...
 
 If a job has optional args, they must be listed so they are passed to the job, in case they exist. The job is responsible for using the optional args or not. (Note: "optional" here is not the same as sequence-level optional args.)
 
-`sets:` specifies the job args that the job sets. The RM checks this. In the example above, the job sets "app", "env", and "node" in `jobArgs`. After calling the job's `Create` method, the RM checks that all three are set in `jobArgs` (with any value, including nil). Like `args:`, this is strict but makes it possible to follow every arg through different sequences. It also makes it explicit which jobs set which args.
+`sets:` specifies the job args that the job sets. The RM checks this. `arg:` and `as:` are to `sets:` as `expected:` and `given:` are to `args:` above: `arg:` is the job arg name set by the job, and `as:` is the job arg name in the specs to use. Even if `arg == as`, both must still be specified.
```
felixplajer, line 106: My fault because I probably forgot to mention this :facepalm:, but can we make the `as` field optional? i.e. if you only put `args`, we assume `as` is the same.

request-manager/grapher/grapher.go
```diff
@@ -630,19 +630,20 @@ func (o *Grapher) remapNodeArgs(n *NodeSpec, args map[string]interface{}) (map[s
...
 		return nil
 	}
 	for _, key := range n.Sets {
 		var ok bool
 		var val interface{}
-		val, ok = argsFrom[key]
+		val, ok = argsFrom[key.Arg]
 		if !ok {
-			return fmt.Errorf("expected %s to set %s in jobargs", n.NodeType, key)
+			return fmt.Errorf("expected %s to set %s as %s in jobargs", n.NodeType, key.Arg, key.As)
```
felixplajer, line 644: I would omit the `as %s` part of the error . If I see `expected job/type to set argA as argB in jobargs`, it's not clear to me whether the job was supposed to set `argA` or `argB`.

util/reformat-yaml/reformat.go
```diff
@@ -0,0 +1,120 @@
+package main
```
felixplajer, line 1: I can't comment on that actual file, but you shouldn't check in the actual `reformat` binary — it won't work on every machine (eg on different OSes). Anyone who wants to run it should build it themselves.

util/reformat-yaml/reformat.go
```diff
@@ -0,0 +1,120 @@
...
+
```
felixplajer, line 2: Can you add a `README.md` file within the reformat-yaml directory, which gives a brief explanation of what this script does (i.e. reformats specs to work with v2 syntax) and how to use it?

docs/v2.0/develop/requests.md
laureneliu: What's the reasoning behind making it optional here, but not for args/given/expected?

felixplajer: We should have done it this way for expected/given too — it's a pain to write both all the time, and most of the time we don't rename args anyways. We should make `given` optional in the future as well.

util/reformat-yaml/test/test.yaml
```diff
@@ -0,0 +1,105 @@
+---
```
felixplajer, line 1: can you provide a file with the expected new yaml from reformat as well?

util/reformat-yaml/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+```
+usage: reformat [input file path] [output file path]
+```
+
+`run.sh` is an example bash script that finds all request specs (files that end with extension `.yaml`) in the Spin Cycle repo and reformats them. Note that `reformat` cannot write to the same file it is reformatting, i.e. the input and output file paths cannot be the same. Thus, `run.sh` writes the reformatted file to a temporary file, then copies that file to the appropriate location.
+
+## Why is reformatting necessary?
+
+See the [release notes](https://github.com/square/spincycle/blob/master/docs/release-notes/index.md).
```
felixplajer, line 17: Link to the doc site instead: https://square.github.io/spincycle/release-notes

Issue #112, opened 2019-06-05
Improve error message "Job Args are missing for <nodes>"


daniel-nichter:
Improve this error message in grapher.go:
```go
return nil, fmt.Errorf("Job Args are missing for %v", missingArgsNodes)
```
It means that a node needed job arg X but no job set X. For example, spec has

```yaml
          - expected: lock
            given: lock
```
That's a typo, should be `given: lockTTL`.

PR #148, opened 2020-07-08
Improve error message when job args are missing

laureneliu:
Issue #112.

### DIFF
--- a/request-manager/grapher/grapher.go
+++ b/request-manager/grapher/grapher.go
@@ -251,7 +251,11 @@ func (o *Grapher) buildComponent(name string, nodeDefs map[string]*NodeSpec, nod
 
 			// First check that all arguments required for the node are present.
 			// If not all arguments are present, then find another node to construct
-			if !o.allArgsPresent(n, nodeArgs) {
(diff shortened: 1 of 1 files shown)


request-manager/grapher/grapher.go
```diff
@@ -410,9 +410,16 @@ func (o *Grapher) buildComponent(name string, nodeDefs map[string]*NodeSpec, nod
 		if nodesCount == len(nodesToBeDone) {
 			missingArgsNodes := []string{}
 			for n, _ := range nodesToBeDone {
-				missingArgsNodes = append(missingArgsNodes, n.Name)
+				// Figure out which args are missing
+				missingArgs := []string{}
+				for _, arg := range n.Args {
+					if _, ok := nodeArgs[arg.Expected]; !ok {
```
felixplajer, line 416: You should be checking arg.Given, not arg.Expected. Also, looking at [allArgsPresent](https://github.com/square/spincycle/blob/c3b603cc7bbcf5a6552f1bc0ba345a2091b2e66c/request-manager/grapher/grapher.go#L582), I think you also need to look at `each` and `if` for missing args.

PR #151, opened 2020-07-21
spinc linter

laureneliu:
SpecParser portion + integration into RM


request-manager/spec/parser.go
```diff
@@ -0,0 +1,132 @@
...
+
+		spec, err, warning := ParseSpec(path)
+		if err != nil {
+			return fmt.Errorf("error reading spec file %s: %s", info.Name(), err)
+		}
+		if warning != nil {
+			printf("Warning: %s: %s\n", info.Name(), warning.Error())
+		}
+		for name, spec := range spec.Sequences {
+			specs.Sequences[name] = spec
```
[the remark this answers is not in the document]
laureneliu: I can add that as a check, but I'm not sure about adding it there. I'm reluctant to break the established SequenceCheck pattern. I do agree that it's a useful check, though.

request-manager/spec/node-check.go
```diff
@@ -0,0 +1,498 @@
...
+	}
+
+	values := map[string]bool{}
+	for _, each := range node.Each {
+		p := strings.Split(each, ":")
+		if len(p) != 2 {
+			continue
+		}
+		if alias, ok := aliases[p[0]]; ok && alias != p[1] {
+			values[p[0]] = true
```
felixplajer, line 229: So, this check does allow
```
each: foo:bar
args:
  - expected: bar
    given: foo
```
but doesn't allow
```
each: foo:bar
args:
  - expected: baz
    given: foo
```
I get that this is because there's already another check which looks at whether each alias == arg expected. However I think this check _should_ still error in the first case — the "each arg == arg given" issue being checked here is a separate issue from the "each alias == arg expected".

request-manager/spec/node-check.go
```diff
@@ -0,0 +1,498 @@
...
+		}
+		aliases[p[0]] = p[1]
+	}
+
+	if len(values) > 0 {
+		return DuplicateValueError{sequenceName, &node.Name, "each", stringSetToArray(values), "in `each: arg:alias`, a given `as` value should not be repeated in `args -> given`; note that if `given` is not explicitly specified, then its value is the same as `expected`"}
```
felixplajer, line 235: there's an `as` in the explanation which I think should be `arg`

request-manager/spec/node-check.go
```diff
@@ -0,0 +1,498 @@
...
+	}
+
+	return nil
+}
+
```
felixplajer, line 240: After looking through all the checks related to `each`, I think maybe we should change terminology: instead of `each: arg:alias` do `each: iterable:iterator` (or `each: list:item` perhaps, I'm open to other suggestions).

Using `arg` in the first one is overloading the meaning of "arg", which is already used when talking about the `args` section of the node.

request-manager/spec/node-check.go
```diff
@@ -0,0 +1,498 @@
...
+		if nodeSet.Arg == nil {
+			return MissingValueError{sequenceName, &node.Name, "sets -> arg", "must be specified for all `sets`"}
+		}
+	}
+
+	return nil
+}
+
+/* ========================================================================== */
+type SetsOnceNodeCheck struct{}
```
felixplajer, line 256: `SetsAsUnique`, admittedly the "as" is not great

request-manager/spec/node-check.go
```diff
@@ -0,0 +1,498 @@
...
+
+	if len(values) > 0 {
+		return DuplicateValueError{sequenceName, &node.Name, "sets -> as", stringSetToArray(values), "note that if `as` is not explicitly specified, then its value is the same as `args`"}
+	}
+
+	return nil
+}
+
+/* ========================================================================== */
+type SetsNotRenamedNodeCheck struct{}
```
felixplajer, line 280: `SetsArgUnique`

request-manager/spec/node-check.go
```diff
@@ -0,0 +1,498 @@
...
+
+	return nil
+}
+
+/* ========================================================================== */
+type EachIfParallelNodeCheck struct{}
+
+/* If `parallel` is set, `each` must be set. */
+func (check EachIfParallelNodeCheck) CheckNode(sequenceName string, node NodeSpec) error {
+	/* `parallel` > 0. */
```
felixplajer, line 308: I'm not sure what this comment is trying to tell me.

request-manager/spec/node-check.go
```diff
@@ -0,0 +1,498 @@
...
+	return nil
+}
+
+/* ========================================================================== */
+type ConditionalHasEqNodeCheck struct{}
+
+/* Conditional nodes must specify `eq`. */
+func (check ConditionalHasEqNodeCheck) CheckNode(sequenceName string, node NodeSpec) error {
+	if node.IsConditional() {
+		if len(node.Eq) == 0 {
```
felixplajer, line 366: Similar to my comment about the `nodes` field in a sequence, if the node _has_ the `eq` field, but it's empty, does that cause a parsing error or result in an empty list? If the second, this field isn't really missing. eg:
```
if: foo
eq:
  # no values here
```

request-manager/spec/node-check.go
```diff
@@ -0,0 +1,498 @@
...
+		if _, err := time.ParseDuration(node.RetryWait); err != nil {
+			return InvalidValueError{sequenceName, &node.Name, "retryWait", []string{node.RetryWait}, "valid duration string"}
+		}
+	}
+
+	return nil
+}
+
+/* ========================================================================== */
+type RequiredArgsProvidedNodeCheck struct {
```
felixplajer, line 444: Can you add a check to make sure that any sequences called actually exist? It should be pretty similar to this one, except at the beginning filter out jobs by checking that node category != job.

request-manager/spec/node-check.go
```diff
@@ -0,0 +1,498 @@
...
+				continue
+			}
+			if _, ok = declaredArgs[*reqArg.Name]; !ok {
+				values[*reqArg.Name] = true
+			}
+		}
+	}
+
+	if len(values) > 0 {
+		explanation := fmt.Sprintf("required arg(s) \"%s\" not given", strings.Join(stringSetToArray(values), ", "))
```
felixplajer, line 493: It's not clear from this message that these are required args for a sequence called by this node — can you fix that? Extra points it you can specify which sequence it is that's missing an arg for conditional nodes which can call multiple sequences.

request-manager/spec/sequence-check.go
```diff
@@ -0,0 +1,161 @@
...
+func (check HasNodesSequenceCheck) CheckSequence(sequence SequenceSpec) error {
+	if len(sequence.Nodes) == 0 {
+		return MissingValueError{sequence.Name, nil, "nodes", "at least one node required"}
+	}
+
+	return nil
+}
+
+/* ========================================================================== */
+type AdminXorOpsSequenceCheck struct{}
```
felixplajer, line 115: put ACL in the name of this somehow, maybe `ACLAdminXorOps`

request-manager/spec/sequence-check.go
```diff
@@ -0,0 +1,161 @@
...
+		if acl.Admin && len(acl.Ops) != 0 {
+			return InvalidValueError{sequence.Name, nil, "acl -> admin", []string{"true"}, "admin=false; alternatively, remove ops"}
+		}
+	}
+
+	return nil
+}
+
+/* ========================================================================== */
+type AclsHaveRolesSequenceCheck struct{}
```
felixplajer, line 129: since `ACL` is an acronym, it should be in all caps in names like this, i.e. `ACLsHaveRoles...`

felixplajer: fyi, when dealing with unexported things, you'd then make all the letters lowercase, like `aclUnexportedThing` rather than `aCLUnexportedThing`. See https://github.com/golang/go/wiki/CodeReviewComments#initialisms

request-manager/spec/sequence-check.go
```diff
@@ -0,0 +1,161 @@
...
+
+/* ACLs must specify a role. */
+func (check AclsHaveRolesSequenceCheck) CheckSequence(sequence SequenceSpec) error {
+	for _, acl := range sequence.ACL {
+		if acl.Role == "" {
```
felixplajer, line 134: makes me wonder, should this be a pointer?

request-manager/spec/common_test.go
```diff
@@ -0,0 +1,126 @@
...
+		for _, node := range sequence.Nodes {
+			if err := check.CheckNode(sequence.Name, *node); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func testFailInvalidValueError(t *testing.T, err error, nodeLevel bool, field, expectedVal, errMsg string) {
```
felixplajer, line 50: rather than passing in a bunch of stuff, could you just pass in the expected error itself? Same applies to next two funcs.

felixplajer: I also don't love that these funcs require the nodes / sequences to be named "node-a" + "seq-a" — it's easy for someone writing a new test not to know that. Passing in the error would help with that.

request-manager/spec/common_test.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	for _, sequence := range allSpecs.Sequences {
+		if err := check.CheckSequence(*sequence); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func runNodeCheckOn(check NodeCheck, filename string) error {
```
felixplajer, line 34: Why don't you just manually create a NodeSpec struct for each test, instead of parsing in a file? That has the benefit of not relying on ParseSpec and of putting the example you're testing right next to the test, which makes it easier to see what exactly is being tested. It's also more "unit test"-like.

The one downside is that you won't be testing whether what you expect in the yaml actually corresponds to what you're writing in the NodeSpec struct. However in most cases the conversion from yaml to struct is obvious, and we can assume the yaml parser acts correctly. It's a good idea to also run some larger "integration" tests which go from yaml through parsing and checking (both static + graph). However I think that can happen later in a different place.

request-manager/spec/make-checks.go
```diff
@@ -0,0 +1,64 @@
+// Copyright 2020, Square, Inc.
+
+package spec
+
+/* ========================================================================== */
+/* Add new checks here. Order shouldn't matter. */
+
+func makeSequenceErrorChecks() []SequenceCheck {
+	return []SequenceCheck{
+		RequiredArgsNamedSequenceCheck{},
```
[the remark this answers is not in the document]
laureneliu: I did it this way to make it explicit what type of check it was to preserve my own sanity. But if you get it wrong (and put it in the wrong list), the compiler should yell at you, so maybe that's not a useful distinction to make in the naming. I'm not super attached to the names. Remove Y/N?

request-manager/spec/sequence-check.go
```diff
@@ -0,0 +1,161 @@
...
+
+	return nil
+}
+
+/* ========================================================================== */
+type HasNodesSequenceCheck struct{}
+
+/* Sequences must contain nodes i.e. are not empty. */
+func (check HasNodesSequenceCheck) CheckSequence(sequence SequenceSpec) error {
+	if len(sequence.Nodes) == 0 {
```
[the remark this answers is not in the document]
laureneliu: It's the latter. To be strictly correct, I'd have to have two checks that distinguish between the field missing, and the field being empty (`MissingValueError` and `InvalidValueError`, respectively). But this feels a bit pedantic to me. If it does stay as one check, I thought it was better as a `MissingValueError`, or at least that it wasn't necessarily better as an `InvalidValueError`.

request-manager/spec/errors.go
```diff
@@ -0,0 +1,111 @@
...
+	case nil:
+		loc = fmt.Sprintf("sequence %s", e.Sequence)
+	default:
+		loc = fmt.Sprintf("sequence %s, node %s", e.Sequence, *e.Node)
+	}
+
+	var values string
+	values = fmt.Sprintf("\"%s\"", strings.Join(e.Values, "\", \""))
+
+	return fmt.Sprintf("%s: invalid value(s) %s in field `%s`, expected %s",
```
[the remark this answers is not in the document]
laureneliu: Using single quotes makes it easier, or using backticks makes it easier? My personal preference is backticks, which is why I used them, but if single quotes are better for more reasons than just personal preference, I can do that.

request-manager/spec/node-check.go
```diff
@@ -0,0 +1,498 @@
...
+}
+
+/* ========================================================================== */
+type ValidEachNodeCheck struct{}
+
+/* `each` values must be in the format `arg:alias`. */
+func (check ValidEachNodeCheck) CheckNode(sequenceName string, node NodeSpec) error {
+	var values []string
+	for _, each := range node.Each {
+		if p := len(strings.Split(each, ":")); p != 2 {
```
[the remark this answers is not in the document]
laureneliu: I have no idea what's up with `p` either; I copied this from grapher. Actually, I'm curious if you have any guesses why it's just called `p` when everything else is named pretty descriptively.

request-manager/spec/node-check.go
```diff
@@ -0,0 +1,498 @@
...
+func (check ArgsEachNotRenamedNodeCheck) CheckNode(sequenceName string, node NodeSpec) error {
+	aliases := map[string]string{}
+	for _, nodeSet := range node.Args {
+		if nodeSet.Given == nil || nodeSet.Expected == nil {
+			continue
+		}
+		// This is fine:
+		// each: [arg:alias]
+		// args:
+		//   - expected: arg
```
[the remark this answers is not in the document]
laureneliu: Well, one reason is that the second case appears in either the docs or our actual specs. (I don't remember right now which it is.)

But also, the second case isn't really renaming the arg--the operation it performs is something like `nodes --> nodes -->> node` (where `-->` is the identity operation, and `-->>` is my attempt to represent a "split" operation). The first case does `nodes --> nodeList` _and_ `nodes -->> node`. So they really are different operations. Whether or not we want to warn in the second case is another matter, but even if we do choose to warn, it'd probably be best as a different check.

request-manager/spec/node-check.go
```diff
@@ -0,0 +1,498 @@
...
+
+	if len(values) > 0 {
+		return DuplicateValueError{sequenceName, &node.Name, "each", stringSetToArray(values), "in `each: arg:alias`, a given `alias` should appear at most once per node"}
+	}
+
+	return nil
+}
+
+/* ========================================================================== */
+type EachNotRenamedNodeCheck struct{}
```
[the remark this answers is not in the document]
laureneliu: Well, it's not really checking that args are unique--this check allows you to repeat an arg as many times as you want, as long as you never rename it.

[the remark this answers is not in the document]
laureneliu: Ah, actually, you mention on a similar comment for `ArgsNotRenamed` that you're fine with it being technically incorrect, so I guess I'll just do this.

request-manager/spec/sequence-check.go
laureneliu: This is how the old code does it (old as in before I started working on it) so I left it as is, since it implies that the empty string is in fact invalid. Of course, then I can't distinguish between an invalid value (empty string) and actually missing (nil).

This is the same problem as above with `sequence.nodes` and something else (was it `each`?) being empty vs actually missing from the yaml entirely. Is it really useful to have separate checks to distinguish between them? I lean towards no, and either creating some new error that says it's "empty or missing", or just sticking with an existing error and clarifying, e.g. having the `MissingValueError` here say "value cannot be the empty string".

request-manager/spec/node-check.go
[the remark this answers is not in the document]
felixplajer: `the second case appears in either the docs or our actual specs` oop. I think there used to be a bug where you (only sometimes, weirdly) had to include the list arg in the `args`, so perhaps it's a relic of that?

In the first case, the result is that the job gets both `node` and `nodeList` (which started as `nodes`) passed in as args. In the second case, it gets both `node` and `nodes` as an arg. Other than the name of the "list" arg, there doesn't really seem to be a difference to me. Using your terms, the second case basically does `nodes --> nodes` (a no-op) and `nodes -->> node`.

I agree in the second case it's not exactly _renaming_ the arg (and in the first case it's more `args`, rather than `each` that's doing "renaming", hence I don't like `ArgsNotRenamedByEach`), but I think the solution is to change the name of the check.

felixplajer:
I didn't exhaustively comment the template grapher / job chain creator — I think in general it would be good to do a read-through of those with an eye to "would this be easy to understand if I didn't already know what it was supposed to do?", and then add comments or break up functions to make the code more readable/friendly to a newcomer. However the actual logic of what those components are doing looks sound!

request-manager/graph/graph.go
```diff
@@ -91,6 +82,94 @@ func (g *Graph) IsValidGraph() bool {
...
+			return false
+		}
+	}
+
+	return true
+}
+
 // Prints out g in DOT graph format.
 // Copy and paste output into http://www.webgraphviz.com/
 func (g *Graph) PrintDot() {
```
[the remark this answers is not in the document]
laureneliu: This is how I write/debug all my graph-related tests lmao

I am quite strongly in favor of keeping it.

request-manager/template/graph.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// graph prety printer function already prints relevant info; there's nothing to add here
+func (g *Node) String() string {
+	return ""
+}
+
+// Template graph (containing template nodes). Includes graph itself, some other metainfo,
+// and useful functions.
+// Graph will have a single source and a single sink node. Graph modification functions enforce this.
```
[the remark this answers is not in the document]
laureneliu: I agree that it's not a lot of info, but it seems to me like that info is more useful in `Grapher` than here, and I don't want to repeat that info again. (I've updated the `Grapher` comment according to earlier feedback, too.) If I were reading this and wanted to know what the struct is used for, I'd do a `grep`, and I think that would give me sufficient info.

I'm going to leave the comment in the code as is and mark this resolved, so this is just documentation of why I made the decision I did.

request-manager/spec/parser.go
```diff
@@ -26,112 +24,95 @@ func ParseSpec(specFile string, logFunc func(string, ...interface{})) (Specs, er
...
-
-	for _, sequence := range allSpecs.Sequences {
-		for _, sequenceCheck := range sequenceErrors {
-			if err := sequenceCheck.CheckSequence(*sequence); err != nil {
-				logFunc("Error: %s\n", err)
-				ret = checkFailed
-			}
-		}
+// Specs require some processing after we've loaded them, but before we run the checker on them.
+func ProcessSpecs(specs Specs) error {
```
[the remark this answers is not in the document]
laureneliu: I think it'd be better to take a `*Specs` and return nothing, since I don't really want to copy all the specs. I mean I could, and it probably won't take noticeably longer, and it only happens on startup, but like why.

It's not clear to me how strongly you prefer things to be purely functional when possible, so I'm not sure which change you think is better.

request-manager/spec/check-factory.go
```diff
@@ -0,0 +1,118 @@
...
+type DefaultCheckFactory struct {
+	AllSpecs Specs // All specs in specs dir
+}
+
+func (c DefaultCheckFactory) MakeSequenceErrorChecks() ([]SequenceCheck, error) {
+	return []SequenceCheck{
+		NoDuplicateArgsSequenceCheck{},
+		RequiredArgsHaveNoDefaultsSequenceCheck{},
+		HasNodesSequenceCheck{},
+		NodesSetsUniqueSequenceCheck{},
```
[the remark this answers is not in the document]
laureneliu: I'm not sure which one you're referring to, since github shows the comment as referring to all four sequence error checks.

request-manager/spec/parser.go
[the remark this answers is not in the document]
felixplajer: I don't have a super strong opinion in this case — this was mostly motivated by our recent reading of Clean Code, which basically says to return `Specs` in this case for ease of reading — but I'm not 100% sold on that so I'm ok with either option. If you do take a `*Specs` and don't return anything, I would probably mention in the comment that the func is modifying the specs passed in.

request-manager/spec/check-factory.go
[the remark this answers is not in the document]
felixplajer: Ah, I meant for `NodesSetsUnique`. Generally (unless a multiline comment) it's the last line before the comment that the comment is referring to.

request-manager/spec/parser.go
```diff
@@ -0,0 +1,118 @@
...
+
+	/* Emit warning if unexpected or duplicate fields are present. */
+	/* Error if specs are incorrectly formatted or fields are of incorrect type. */
+	warn := yaml.UnmarshalStrict(sequenceData, &spec)
+	if warn != nil {
+		err = yaml.Unmarshal(sequenceData, &spec)
+		if err != nil {
+			return spec, err
+		}
+		logFunc("Warning: %s", warn)
```
felixplajer, line 33: Can you log the file name here? Otherwise you can't tell which file a warning is coming from.

request-manager/spec/check-factory.go
```diff
@@ -0,0 +1,118 @@
...
+
+		NoExtraSequenceArgsProvidedNodeCheck{c.AllSpecs},
+	}, nil
+}
+
+func (c DefaultCheckFactory) MakeNodeWarningChecks() ([]NodeCheck, error) {
+	return []NodeCheck{
+		EachNotRenamedTwiceNodeCheck{},
+		ArgsNotRenamedTwiceNodeCheck{},
+		EachElementDoesNotDuplicateArgsExpectedNodeCheck{},
```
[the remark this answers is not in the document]
laureneliu: This was supposed to be an error--looks like I had a had a copy/paste error and added that instead of `EachListDoesNotDuplicateArgsGivenNodeCheck` as a warning.

In that case, it just gets overwritten [here](https://github.com/square/spincycle/blob/spinc-linter/request-manager/chain/creator.go#L235-L242). I'm inclined to keep it as a default check, because it won't actually break the RM. Base checks currently enforce assumptions in template/chain, and if any assumption is false, template/chain segfaults.

request-manager/spec/node-check.go
```diff
@@ -35,11 +35,15 @@ type ValidCategoryNodeCheck struct{}
 /* 'category: (job | sequence | conditional)' */
 func (check ValidCategoryNodeCheck) CheckNode(sequenceName string, node Node) error {
 	if !node.IsJob() && !node.IsSequence() && !node.IsConditional() {
+		var category string
+		if node.Category != nil {
```
felixplajer, line 39: You should check if category == nil before the is job/sequence/conditional checks. Else if category is not set, you'll get 2 errors — from `HasCategoryNodeCheck` and from this check.

request-manager/spec/node-check.go
```diff
@@ -762,3 +732,37 @@ func (check SubsequencesExistNodeCheck) CheckNode(sequenceName string, node Node
...
 }
+
+/* ========================================================================== */
+// Helper functions
+
+// Get list of all sequences called by node
+func getCalledSequences(node Node) []string {
+	var sequences []string
+	if node.IsSequence() {
+		sequences = []string{*node.NodeType}
```
felixplajer, line 743: beep boop, gotta check to make sure NodeType isn't nil

request-manager/spec/parser.go
```diff
@@ -0,0 +1,119 @@
...
+func ProcessSpecs(specs *Specs) error {
+	for sequenceName, sequence := range specs.Sequences {
+		sequence.Name = sequenceName
+
+		for nodeName, node := range sequence.Nodes {
+			node.Name = nodeName
+
+			// Set various optional fields if they were excluded.
+			for i, nodeSet := range node.Sets {
+				if nodeSet.As == nil {
```
felixplajer, line 103: Before doing this, you actually also need to check that `nodeSet != nil`, since node.Sets is a `[]*NodeSet`

request-manager/spec/parser.go
```diff
@@ -90,7 +90,8 @@ func ParseSpecsDir(specsDir string, logFunc func(string, ...interface{})) (Specs
 }
 
 // Specs require some processing after we've loaded them, but before we run the checker on them.
-func ProcessSpecs(specs Specs) error {
+// Function modifies specs passed in.
+func ProcessSpecs(specs *Specs) error {
```
felixplajer, line 94: I think we discussed removing the `error` return here?

request-manager/template/grapher.go
```diff
@@ -228,31 +230,34 @@ func (o *Grapher) getTemplate(seq *spec.Sequence) (*Graph, error) {
 	jobArgs := getAllSequenceArgs(seq)
 
 	// Create a graph node for every node in the spec
-	components := map[*spec.Node]*Node{}
+	// Key on node names--they should be unique within a sequence (otherwise, dependencies
```
felixplajer, line 233: err, is there a check that node names are unique within a sequence?

laureneliu: The yaml parser errors if that happens. I don't remember the exact error message, but I think it was sufficiently descriptive? Either way, I seem to remember that there's no good way to distinguish between different types of errors thrown by the parser.

felixplajer: Ah neat, that works.

linter/app/app.go
```diff
@@ -20,21 +20,29 @@ type Context struct {
...
 	GeneratorFactory id.GeneratorFactory
-	CheckFactories   []spec.CheckFactory // All additional check factories to run
 }
 
 type Hooks struct {
+	// Parse specs
 	LoadSpecs func(specsDir string, logFunc func(string, ...interface{})) (spec.Specs, error)
+
+	// Get additional checks to run
+	GetCheckFactories func(specs spec.Specs) []spec.CheckFactory
```
felixplajer, line 30: What if instead of directly storing various factories, the `Factories` struct has 2 funcs — `MakeIdGeneratorFactory` and `MakeCheckFactories` (and that second one can take in the specs as an argument). That's similar to what the RM does, and I think makes a bit more sense — when I think of a "hook" I expect it to _do_ something, not just create a factory.

PR #155, opened 2020-09-01
Improve spinc-linter code

laureneliu:


daniel-nichter:
For the record, we're putting all graph-related code back in a single `graph/` pkg. Although we have different uses of graphs, under the hood it's all graph and graph logic, so one graph pkg makes sense and makes it a lot easier to following graphing logic which is all closely related.

PR #156, opened 2020-09-02
Rename graph code variables+components

laureneliu:
Part 2/? of spinc linter improvements

### DIFF
--- a/request-manager/graph/graph.go
+++ b/request-manager/graph/graph.go
@@ -60,7 +60,7 @@ func (g *Graph) IsValidGraph() error {
 	return nil
 }
 
-// GetNext returns the out edges of a node (node id --> Node)
+// GetNext returns the out edges of a node
 func (g *Graph) GetNext(n *Node) []*Node {
 	next := make([]*Node, len(g.Edges[n.Id]))
 	for i, nextId := range g.Edges[n.Id] {
@@ -69,7 +69,7 @@ func (g *Graph) GetNext(n *Node) []*Node {
 	return next
 }
 
-// GetPrev returns the in edges of a node (node id --> Node)
+// GetPrev returns the in edges of a node
 func (g *Graph) GetPrev(n *Node) []*Node {
 	prev := make([]*Node, len(g.RevEdges[n.Id]))
 	for i, prevId := range g.RevEdges[n.Id] {
--- a/request-manager/graph/graph.go
+++ b/request-manager/graph/graph.go
@@ -5,7 +5,6 @@ package graph
 import (
 	"fmt"
 
-	"github.com/square/spincycle/v2/job"
 	"github.com/square/spincycle/v2/request-manager/spec"
 )
 
@@ -36,7 +35,7 @@ type Node struct {
 	Spec *spec.Node // Node spec that this graph node represents
 
 	// Used when node represents a job
-	Job               job.Job                // Job that this graph node represents
+	JobBytes          []byte                 // return value of Job.Serialize method
 	Args              map[string]interface{} // The args the node was created with
 	Retry             uint                   // The number of times to retry a node
 	RetryWait         string                 // The time to sleep between retries
@@ -48,24 +47,33 @@ type Node struct {
 // IsValidGraph asserts that g is a valid graph by ensuring that
 // g is acyclic and connected (not fully connected), and that the edge
 // map matches the reverse edge map.
-func (g *Graph) IsValidGraph() bool {
-	return !g.hasCycles() && g.isConnected() && g.edgesMatchesRevEdges()
+func (g *Graph) IsValidGraph() error {
+	if g.hasCycles() {
+		return fmt.Errorf("cycles found in graph")
+	}
+	if !g.isConnected() {
+		return fmt.Errorf("graph not connected")
+	}
+	if !g.edgesMatchesRevEdges() {
+		return fmt.Errorf("adjacency list does not match reverse adjacency list")
+	}
+	return nil
 }
 
 // GetNext returns the out edges of a node (node id --> Node)
-func (g *Graph) GetNext(n *Node) map[string]*Node {
-	next := map[string]*Node{}
-	for _, nextId := range g.Edges[n.Id] {
-		next[nextId] = g.Nodes[nextId]
+func (g *Graph) GetNext(n *Node) []*Node {
+	next := make([]*Node, len(g.Edges[n.Id]))
+	for i, nextId := range g.Edges[n.Id] {
+		next[i] = g.Nodes[nextId]
 	}
 	return next
 }
 
 // GetPrev returns the in edges of a node (node id --> Node)
-func (g *Graph) GetPrev(n *Node) map[string]*Node {
-	prev := map[string]*Node{}
-	for _, prevId := range g.RevEdges[n.Id] {
-		prev[prevId] = g.Nodes[prevId]
+func (g *Graph) GetPrev(n *Node) []*Node {
+	prev := make([]*Node, len(g.RevEdges[n.Id]))
+	for i, prevId := range g.RevEdges[n.Id] {
+		prev[i] = g.Nodes[prevId]
 	}
 	return prev
 }
@@ -77,8 +85,11 @@ func (g *Graph) GetPrev(n *Node) map[string]*Node {
 //      prev and next both are present in g
 //      next "comes after" prev in the graph, when traversing from the source node
 func (g *Graph) InsertComponentBetween(component *Graph, prev *Node, next *Node) error {
-	if !g.IsValidGraph() || !component.IsValidGraph() {
-		return fmt.Errorf("Graph not valid!")
+	if err := g.IsValidGraph(); err != nil {
+		return fmt.Errorf("graph to insert component into: %s", err)
+	}
+	if err := component.IsValidGraph(); err != nil {
+		return fmt.Errorf("component to be inserted: %s", err)
 	}
 
 	var source, sink string
@@ -129,8 +140,8 @@ func (g *Graph) InsertComponentBetween(component *Graph, prev *Node, next *Node)
 	}
 
 	// Verify resulting graph is ok
-	if !g.IsValidGraph() {
-		return fmt.Errorf("graph not valid after insert")
+	if err := g.IsValidGraph(); err != nil {
+		return fmt.Errorf("graph not valid after insert: %s", err)
 	}
 	return nil
 }
@@ -163,18 +174,18 @@ func (g *Graph) PrintDot() {
 // hasCycles returns true iff the graph has at least one cycle in it.
 func (g *Graph) hasCycles() bool {
 	seen := map[string]*Node{g.Source.Id: g.Source}
-	return g.hasCyclesDFS(seen, g.Source)
+	return g.hasCyclesHelper(seen, g.Source)
 }
 
 // isConnected returns true iff every node is reachable from the start node, and
 // every path terminates at the end node.
 func (g *Graph) isConnected() bool {
 	// Check forwards connectivity and backwards connectivity
-	return g.connectedToLastNodeDFS(g.Source) && g.connectedToFirstNodeDFS(g.Sink)
+	return g.connectedToLastNode(g.Source) && g.connectedToFirstNode(g.Sink)
 }
 
-// connectedToLastNodeDFS returns true iff the last node in g is reachable from n.
-func (g *Graph) connectedToLastNodeDFS(n *Node) bool {
+// connectedToLastNode returns true iff the last node in g is reachable from n.
+func (g *Graph) connectedToLastNode(n *Node) bool {
 	if n == nil {
 		return false
 	}
@@ -187,7 +198,7 @@ func (g *Graph) connectedToLastNodeDFS(n *Node) bool {
 	for _, next := range g.GetNext(n) {
 
 		// Every node after n must also be connected to the last node
-		connected := g.connectedToLastNodeDFS(next)
+		connected := g.connectedToLastNode(next)
 		if !connected {
 			return false
 		}
@@ -195,8 +206,8 @@ func (g *Graph) connectedToLastNodeDFS(n *Node) bool {
 	return true
 }
 
-// connectedToFirstNodeDFS returns true if n is reachable from the first node in g.
-func (g *Graph) connectedToFirstNodeDFS(n *Node) bool {
+// connectedToFirstNode returns true if n is reachable from the first node in g.
+func (g *Graph) connectedToFirstNode(n *Node) bool {
 	if n == nil {
 		return false
 	}
@@ -209,7 +220,7 @@ func (g *Graph) connectedToFirstNodeDFS(n *Node) bool {
 	for _, prev := range g.GetPrev(n) {
 
 		// Every node before n must also be connected to the first node
-		connected := g.connectedToFirstNodeDFS(prev)
+		connected := g.connectedToFirstNode(prev)
 		if !connected {
 			return false
 		}
@@ -239,9 +250,9 @@ func (g *Graph) edgesMatchesRevEdges() bool {
 	return true
 }
 
-// hasCyclesDFS determines if a graph has cycles, using DFS.
+// hasCyclesHelper determines if a graph has cycles, using DFS.
 // Precondition: start node is already in seen list.
-func (g *Graph) hasCyclesDFS(seen map[string]*Node, start *Node) bool {
+func (g *Graph) hasCyclesHelper(seen map[string]*Node, start *Node) bool {
 	for _, next := range g.GetNext(start) {
 
 		// If the next node has already been seen, return true
@@ -253,7 +264,7 @@ func (g *Graph) hasCyclesDFS(seen map[string]*Node, start *Node) bool {
 		seen[next.Id] = next
 
 		// Continue searching after next node
-		if g.hasCyclesDFS(seen, next) {
+		if g.hasCyclesHelper(seen, next) {
 			return true
 		}
 
--- a/request-manager/graph/graph.go
+++ b/request-manager/graph/graph.go
@@ -70,7 +70,8 @@ func (g *Graph) GetPrev(n *Node) map[string]*Node {
 	return prev
 }
 
-// InsertComponentBetween takes a Graph as input and inserts it between the given prev and next nodes.
+// InsertComponentBetween takes a Graph as input and inserts it between the given
+// prev and next nodes.
 // Preconditions:
 //      component and g are connected and acyclic
 //      prev and next both are present in g
@@ -165,8 +166,8 @@ func (g *Graph) hasCycles() bool {
 	return g.hasCyclesDFS(seen, g.Source)
 }
 
-// isConnected returns true iff every node is reachable from the start node, and every path
-// terminates at the end node.
+// isConnected returns true iff every node is reachable from the start node, and
+// every path terminates at the end node.
 func (g *Graph) isConnected() bool {
 	// Check forwards connectivity and backwards connectivity
 	return g.connectedToLastNodeDFS(g.Source) && g.connectedToFirstNodeDFS(g.Sink)
--- a/request-manager/graph/graph.go
+++ b/request-manager/graph/graph.go
@@ -11,12 +11,13 @@ import (
 
 // Graph is a DAG where nodes are node specs (if a sequence graph) or jobs (if a
 // request graph), and a directed edge (u, v) exists iff u is a dependency of v.
+// It has one source and one sink node.
 type Graph struct {
-	Name   string // Name of the Graph
+	Name   string // Name of the graph
 	Source *Node  // The single source node of the graph
 	Sink   *Node  // The single sink node of the graph
 
-	// Nodes are keyed on their node id, which is a string
+	// Nodes are keyed on node id, which is a string
 	Nodes    map[string]*Node    // All nodes
 	Edges    map[string][]string // All edges
 	RevEdges map[string][]string // All reverse edges (sink -> source)
@@ -29,7 +30,7 @@ type Graph struct {
 type Node struct {
 	// Always valid
 	Id   string // UID within graph
-	Name string // Descriptive name for debugging graph or job chain
+	Name string // Descriptive name for debugging and visualizing graph
 
 	// Used when node represents a node spec
 	Spec *spec.Node // Node spec that this graph node represents
@@ -44,14 +45,14 @@ type Node struct {
 	SequenceRetryWait string                 // The time to sleep between sequence retries
 }
 
-// Asserts that g is a valid graph (according to Grapher's use case).
-// Ensures that g is acyclic, is connected (not fully connected),
-// and edge map matches reverse edge map.
+// IsValidGraph asserts that g is a valid graph by ensuring that
+// g is acyclic and connected (not fully connected), and that the edge
+// map matches the reverse edge map.
 func (g *Graph) IsValidGraph() bool {
 	return !g.hasCycles() && g.isConnected() && g.edgesMatchesRevEdges()
 }
 
-// Get out edges of node (node id --> Node)
+// GetNext returns the out edges of a node (node id --> Node)
 func (g *Graph) GetNext(n *Node) map[string]*Node {
 	next := map[string]*Node{}
 	for _, nextId := range g.Edges[n.Id] {
@@ -60,7 +61,7 @@ func (g *Graph) GetNext(n *Node) map[string]*Node {
 	return next
 }
 
-// Get in edges of node (node id --> Node)
+// GetPrev returns the in edges of a node (node id --> Node)
 func (g *Graph) GetPrev(n *Node) map[string]*Node {
 	prev := map[string]*Node{}
 	for _, prevId := range g.RevEdges[n.Id] {
@@ -69,7 +70,7 @@ func (g *Graph) GetPrev(n *Node) map[string]*Node {
 	return prev
 }
 
-// InsertComponentBetween will take a Graph as input, and insert it between the given prev and next nodes.
+// InsertComponentBetween takes a Graph as input and inserts it between the given prev and next nodes.
 // Preconditions:
 //      component and g are connected and acyclic
 //      prev and next both are present in g
@@ -86,7 +87,7 @@ func (g *Graph) InsertComponentBetween(component *Graph, prev *Node, next *Node)
 		g.Nodes[k] = v
 	}
 
-	// update adjacency lists
+	// Update adjacency lists
 	for k, v := range component.Edges {
 		g.Edges[k] = v
 	}
@@ -94,7 +95,7 @@ func (g *Graph) InsertComponentBetween(component *Graph, prev *Node, next *Node)
 		g.RevEdges[k] = v
 	}
 
-	// connect previous node and start node of component
+	// Connect previous node and start node of component
 	source = prev.Id
 	sink = component.Source.Id
 	if find(g.Edges[source], sink) < 0 {
@@ -102,7 +103,7 @@ func (g *Graph) InsertComponentBetween(component *Graph, prev *Node, next *Node)
 		g.RevEdges[sink] = append(g.RevEdges[sink], source)
 	}
 
-	// connect last node of component and next node
+	// Connect last node of component and next node
 	source = component.Sink.Id
 	sink = next.Id
 	if find(g.Edges[source], sink) < 0 {
@@ -110,7 +111,7 @@ func (g *Graph) InsertComponentBetween(component *Graph, prev *Node, next *Node)
 		g.RevEdges[sink] = append(g.RevEdges[sink], source)
 	}
 
-	// remove all connections between prev and next
+	// Remove all connections between prev and next
 	source = prev.Id
 	sink = next.Id
 	i := find(g.Edges[source], sink)
@@ -126,15 +127,16 @@ func (g *Graph) InsertComponentBetween(component *Graph, prev *Node, next *Node)
 		i = find(g.Edges[sink], source)
 	}
 
-	// verify resulting graph is ok
+	// Verify resulting graph is ok
 	if !g.IsValidGraph() {
 		return fmt.Errorf("graph not valid after insert")
 	}
 	return nil
 }
 
-// Prints out g in DOT graph format.
+// PrintDot prints out g in DOT graph format.
 // Copy and paste output into http://www.webgraphviz.com/
+// It's not used anywhere in the code, but it is a very useful debugging tool.
 func (g *Graph) PrintDot() {
 	fmt.Printf("digraph {\n")
 	fmt.Printf("\trankdir=UD;\n")
@@ -155,51 +157,22 @@ func (g *Graph) PrintDot() {
 	fmt.Println("}")
 }
 
-// Returns true if a matches b, regardless of ordering
-func SlicesMatch(a, b []string) bool {
-	if a == nil && b == nil {
-		return true
-	}
-
-	if a == nil || b == nil {
-		return false
-	}
-
-	if len(a) != len(b) {
-		return false
-	}
-
-	for i, _ := range a {
-		ok := false
-		for j, _ := range b {
-			if a[i] == b[j] {
-				ok = true
-			}
-		}
-		if !ok {
-			return false
-		}
-	}
-
-	return true
-}
-
 // --------------------------------------------------------------------------
 
-// returns true iff the graph has at least one cycle in it
+// hasCycles returns true iff the graph has at least one cycle in it.
 func (g *Graph) hasCycles() bool {
 	seen := map[string]*Node{g.Source.Id: g.Source}
 	return g.hasCyclesDFS(seen, g.Source)
 }
 
-// returns true iff every node is reachable from the start node, and every path
-// terminates at the end node
+// isConnected returns true iff every node is reachable from the start node, and every path
+// terminates at the end node.
 func (g *Graph) isConnected() bool {
 	// Check forwards connectivity and backwards connectivity
 	return g.connectedToLastNodeDFS(g.Source) && g.connectedToFirstNodeDFS(g.Sink)
 }
 
-// Returns true if the last node in g is reachable from n
+// connectedToLastNodeDFS returns true iff the last node in g is reachable from n.
 func (g *Graph) connectedToLastNodeDFS(n *Node) bool {
 	if n == nil {
 		return false
@@ -221,7 +194,7 @@ func (g *Graph) connectedToLastNodeDFS(n *Node) bool {
 	return true
 }
 
-// Returns true if n is reachable from the first node in g
+// connectedToFirstNodeDFS returns true if n is reachable from the first node in g.
 func (g *Graph) connectedToFirstNodeDFS(n *Node) bool {
 	if n == nil {
 		return false
@@ -243,7 +216,8 @@ func (g *Graph) connectedToFirstNodeDFS(n *Node) bool {
 	return true
 }
 
-// Returns true iff `Edges` represents exactly the same set of edges as `RevEdges`.
+// edgesMatchesRevEdges returns true iff `Edges` represents exactly the same set
+// of edges as `RevEdges`.
 func (g *Graph) edgesMatchesRevEdges() bool {
 	for source, sinks := range g.Edges {
 		for _, sink := range sinks {
@@ -264,8 +238,8 @@ func (g *Graph) edgesMatchesRevEdges() bool {
 	return true
 }
 
-// Determines if a graph has cycles, using dfs
-// precondition: start node is already in seen list
+// hasCyclesDFS determines if a graph has cycles, using DFS.
+// Precondition: start node is already in seen list.
 func (g *Graph) hasCyclesDFS(seen map[string]*Node, start *Node) bool {
 	for _, next := range g.GetNext(start) {
 
@@ -288,7 +262,7 @@ func (g *Graph) hasCyclesDFS(seen map[string]*Node, start *Node) bool {
 	return false
 }
 
-// returns the index of s in ss, returns -1 if s is not found in ss
+// find returns the index of s in ss, returns -1 if s is not found in ss.
 func find(ss []string, s string) int {
 	for i, j := range ss {
 		if j == s {
--- a/request-manager/graph/grapher.go
+++ b/request-manager/graph/grapher.go
@@ -274,7 +274,7 @@ func buildSeqGraph(seqSpec *spec.Sequence, idgen id.Generator) (seqGraph *Graph,
 		// after this loop.
 		for nodeName, node := range nodesToAdd {
 			nodeSpec := seqSpec.Nodes[nodeName]
-			if !dependenciesSatisfied(nodesAdded, nodeSpec.Dependencies) {
+			if !haveAllDeps(nodesAdded, nodeSpec.Dependencies) {
 				continue
 			}
 
@@ -399,9 +399,9 @@ func getAllSequenceArgs(seq *spec.Sequence) map[string]bool {
 	return jobArgs
 }
 
-// dependenciesSatifised checks whether the set of nodes in a graph
+// haveAllDeps checks whether the set of nodes in a graph
 // satisfies all dependencies.
-func dependenciesSatisfied(inGraph map[string]bool, dependencies []string) bool {
+func haveAllDeps(inGraph map[string]bool, dependencies []string) bool {
 	for _, dep := range dependencies {
 		if _, ok := inGraph[dep]; !ok {
 			return false
@@ -412,6 +412,15 @@ func dependenciesSatisfied(inGraph map[string]bool, dependencies []string) bool 
 
 // checkNodeArgs checks whether all node args are present in the job args map.
 func checkNodeArgs(n *spec.Node, jobArgs map[string]bool) error {
+	// If this is a conditional node, assert that the "if" job arg is present
+	// in the job args map.
+	// Static checks assert that for conditional nodes, if != nil.
+	if n.IsConditional() {
+		if !jobArgs[*n.If] {
+			return fmt.Errorf("in node %s: 'if: %s': job arg %s is not set", n.Name, *n.If, *n.If)
+		}
+	}
+
 	missing := []string{}
 
 	// Assert that the iterable variable is present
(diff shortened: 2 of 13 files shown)


request-manager/graph/grapher.go
```diff
@@ -10,17 +10,20 @@ import (
...
 // a node spec, and edges indicate dependencies between nodes in the spec. Each
 // sequence graphs describes a single sequence spec.
-// Sequence, conditional, and expandable sequence nodes are represented by one
-// node in a sequence graph, and are later resolved by the Resolver.
+// Sequence, conditional, and expandable sequence nodes correspond to a single
+// node in a sequence graph, and are later resolved into a request subgraph
+// consisting of only job nodes by the Resolver.
 type Grapher struct {
 	// User-provided
 	allSequences map[string]*spec.Sequence // All sequences read in from request specs
```
daniel-nichter, line 21: Better name for this var: `seqSpecs` to make it clear at a glance that it's specs.

request-manager/graph/grapher.go
```diff
@@ -31,13 +34,16 @@ func NewGrapher(specs spec.Specs, idGenFactory id.GeneratorFactory) *Grapher {
 // DoChecks performs graph checks for all sequences.
 func (gr *Grapher) DoChecks() (seqGraphs map[string]*Graph, seqErrors map[string]error) {
```
daniel-nichter, line 35: Re func name: maybe `CheckSequences` or something. "Do" falls into the same realm as not using "Get" (https://golang.org/doc/effective_go.html#Getters). Also, imho, all functions "do" so it doesn't add anything in terms of clarity, etc.

daniel-nichter: Code comment can say a little more. The import thing about this func is that it returns both the final/official "list" (it's really a map) of sequence graphs _and_ all graph check errors. And also important: are the two returns mutually exclusive? Normally, with `func() (<var>, err)` the two are mutually exclusive: if err != nil then caller must ignore <var>.

request-manager/graph/grapher.go
```diff
@@ -31,13 +34,16 @@ func NewGrapher(specs spec.Specs, idGenFactory id.GeneratorFactory) *Grapher {
...
+
 	// First, build all the sequence graphs.
 	// Since sequence graphs are just a graph representation of dependencies
-	// between a single sequence's nodes, this is per sequence; there's no
-	// need to look at any subsequences.
+	// between a single sequence's nodes, this is contained to one sequence;
+	// this won't look at any subsequences.
 	for seqName, seqSpec := range gr.allSequences {
 		// Generates IDs unique within sequence graph
 		idgen := gr.idGenFactory.Make()
```
daniel-nichter, line 49: I think at this point we do not need to generate IDs? Sequence graphs are (to use the previous term) templates (or prototypes, if you like), so they're generic. It's not until we resolve them into request graphs that they (and job nodes) need unique IDs (unique only within a request graph, not universally unique).

request-manager/graph/grapher.go
```diff
@@ -177,25 +189,32 @@ func getMissingSets(actual map[string]bool, declared []*spec.NodeSet) []string {
...
 	// The graph we'll be filling in
-	seqGraph, err = newGraph(seqSpec.Name, idgen)
+	seqGraph, err = newSeqGraph(seqSpec.Name, idgen)
 	if err != nil {
 		return nil, nil, err
 	}
-	// newGraph doesn't manager `Order` for us; we'll have to maintain it ourselves in this function
+
+	// We have to maintain `Order` ourselves
 	seqGraph.Order = append(seqGraph.Order, seqGraph.Source)
```
daniel-nichter, line 205: I think the next iteration of `type Graph struct` should do this internally. Order is a property of the Graph, so the Graph should maintain it, leaving the caller to simply use the Graph without worrying about internal, low-level graph details.

request-manager/graph/grapher.go
```diff
@@ -177,25 +189,32 @@ func getMissingSets(actual map[string]bool, declared []*spec.NodeSet) []string {
...
 
-	// Set of job args available, i.e. sequence args + args set by nodes in graph so far
+	// Set of job args available, i.e. sequence args + args set by nodes in
+	// graph so far.
+	// We start out with just the sequence args.
 	jobArgs := getAllSequenceArgs(seqSpec)
```
daniel-nichter, line 210: Let's move this var down closer to where/when it's actually used. This is a good practice in general, with one exception being that sometimes we define some widely-used vars at the very top of a func. But otherwise, it helps when reading code that vars appear "just in time".

request-manager/graph/grapher.go
```diff
@@ -177,25 +189,32 @@ func getMissingSets(actual map[string]bool, declared []*spec.NodeSet) []string {
...
 
-	// Create a graph with a single node for every node in the spec
-	// Key on node names--they should be unique within a sequence (otherwise, dependencies
-	// are ill-defined)
+	// Create a graph consisting of a single node for every node in the spec.
+	// There's a function to insert a subgraph into a graph, but not to insert
+	// a node into a graph. Creating these as single-node graphs upfront rather
+	// than nodes makes the code more readable later.
+	// Key on node names; they should be unique within a sequence (otherwise,
+	// dependencies are ill-defined).
```
daniel-nichter, line 217: Pretty good comment, but after a few close reads and reading the actual code block, I think it'd be more clear as:
```
	// Create graph nodes for every node in the sequence spec. We're not connecting
	// the nodes yet (i.e. not building a graph), just initializing all the graph
	// nodes. In the next loop, we'll wire them up (i.e. build the sequence graph).
	//
	// Note: graph nodes can be any type (job, seq, or conditional), but for
	// sequence graphs we don't do anything (i.e. don't create job.Job for job node);
	// a Resolver does that when creating a request graph (RG). Sequence graphs
	// are only (graph) structure and metadata about that structure.
```

request-manager/graph/grapher.go
```diff
@@ -217,25 +236,27 @@ func buildSeqGraph(seqSpec *spec.Sequence, idgen id.Generator) (seqGraph *Graph,
 		}
 	}
 
-	// Nodes we've yet to add
+	// Nodes we've yet to add, represented as single-node graphs
 	nodesToAdd := map[string]*Graph{}
 	for k, v := range nodes {
 		nodesToAdd[k] = v
```
daniel-nichter, line 242: You can do this in the previous loop: i.e. create the nodes and add them to `nodesToAdd`. It's an easy micro-optimization.

request-manager/graph/grapher.go
```diff
@@ -217,25 +236,27 @@ func buildSeqGraph(seqSpec *spec.Sequence, idgen id.Generator) (seqGraph *Graph,
...
 	}
 	nodesAdded := map[string]bool{}
 
 	// Build graph by adding nodes, starting from the source node, and then
-	// adding all adjacent nodes to the source node, and so on.
-	// We cannot add nodes in any order because we do not know the reverse dependencies
+	// adding all adjacent nodes to the source node, and so on
+	// We cannot add nodes in any order because we do not know the reverse
+	// dependencies
```
daniel-nichter, line 249: Good comment again. We can make it a little better. The most important thing for future engineers to know here: the nodes in the request spec are _not_ in topological order, or any order, i.e. spec order != graph order. In spec, graph order is implied by the `deps` field. So here in code, we continuously loop through the (potentially random) spec order, creating nodes if and when all dependent nodes have been created. Usually , spec order is close to topological order, but it's not a rule.

request-manager/graph/grapher.go
```diff
@@ -217,25 +236,27 @@ func buildSeqGraph(seqSpec *spec.Sequence, idgen id.Generator) (seqGraph *Graph,
...
 	for len(nodesToAdd) > 0 {
 
-		componentAdded := false
+		nodeAdded := false
```
daniel-nichter, line 252: Why do we presume this to be false from the start? (I know why, but future eng. might wonder.)

request-manager/graph/grapher.go
```diff
@@ -217,25 +236,27 @@ func buildSeqGraph(seqSpec *spec.Sequence, idgen id.Generator) (seqGraph *Graph,
...
 
-		for nodeName, component := range nodesToAdd {
+		for nodeName, node := range nodesToAdd {
 			nodeSpec := seqSpec.Nodes[nodeName]
 			if dependenciesSatisfied(nodesAdded, nodeSpec.Dependencies) {
```
daniel-nichter, line 256:
```go
if !dependenciesSatisfied(nodesAdded, nodeSpec.Dependencies) {
    continue
}

// rest of the code
```

From https://golang.org/doc/effective_go.html#if:

> This is an example of a common situation where code must guard against a sequence of error conditions. The code reads well if the successful flow of control runs down the page, eliminating error cases as they arise. 

Go code should "read down the left".

request-manager/graph/grapher.go
```diff
@@ -361,11 +386,11 @@ func dependenciesSatisfied(inGraph map[string]bool, dependencies []string) bool
 	return true
 }
 
-// Returns a list of node args that aren't present in the job args map.
+// getMissingArgs returns a list of node args that aren't present in the job args map.
 func getMissingArgs(n *spec.Node, jobArgs map[string]bool) ([]string, error) {
```
daniel-nichter, line 390: This does more than get the missing, it does all the node arg checks, so s/getMissingArgs/checkNodeArgs/ or similar. And since we really only care about error or not, it can return a single `error` and internally, if there are multiple missing args, it can do like, `return fmt.Errorf("missing args %s", strings.Join(missingArgs, ", "))`

request-manager/graph/grapher.go
```diff
@@ -244,26 +265,27 @@ func buildSeqGraph(seqSpec *spec.Sequence, idgen id.Generator) (seqGraph *Graph,
 					return nil, nil, fmt.Errorf("node %s missing job args: %s", nodeSpec.Name, strings.Join(missingArgs, ", "))
 				}
 
-				// Insert component into graph
+				// Insert node into graph
 				if len(nodeSpec.Dependencies) == 0 {
 					// Case: no dependencies; insert directly after start node
-					err := seqGraph.InsertComponentBetween(component, seqGraph.Source, seqGraph.Sink)
+					err := seqGraph.InsertComponentBetween(node, seqGraph.Source, seqGraph.Sink)
```
daniel-nichter, line 271: I know what's happening here and why, but it's part of the code that has always confused me. If a node has zero deps, that means it has zero in-edges, e.g. A in A->B->C. But the confusing part is: what if, in that example, we have already processed C and B, then how is A inserted between `seqGraph.Source, seqGraph.Sink`? Aren't there already 2 nodes (B and C) between source and sink?

request-manager/graph/grapher.go
```diff
@@ -244,26 +265,27 @@ func buildSeqGraph(seqSpec *spec.Sequence, idgen id.Generator) (seqGraph *Graph,
...
 						if err != nil {
 							return nil, nil, err
 						}
 					}
 				}
 
-				// Add to topological ordering
-				seqGraph.Order = append(seqGraph.Order, component.Source)
+				// Add this node to topological ordering
+				seqGraph.Order = append(seqGraph.Order, node.Source)
```
daniel-nichter, line 288: Same comment as above: in next iteration of Graph, I hope to see this kind of low-level graph handling logic hidden at this level of code (i.e. caller of graph shouldn't have to do or know anything about graph's topo order).

request-manager/graph/grapher.go
```diff
@@ -272,13 +294,14 @@ func buildSeqGraph(seqSpec *spec.Sequence, idgen id.Generator) (seqGraph *Graph,
 
 				delete(nodesToAdd, nodeName)
 				nodesAdded[nodeName] = true
-				componentAdded = true
+				nodeAdded = true
```
daniel-nichter, line 297: I don't think this is needed, you can use `length(nodesAdded)` instead?

request-manager/graph/grapher.go
```diff
@@ -272,13 +294,14 @@ func buildSeqGraph(seqSpec *spec.Sequence, idgen id.Generator) (seqGraph *Graph,
...
 			}
 
 		}
 
-		// If we were unable to add nodes, there must be a cyclical dependency, which is an error
-		if !componentAdded {
+		// If we were unable to add nodes, there must be a cyclical
+		// dependency, which is an error
+		if !nodeAdded {
```
daniel-nichter, line 304: Is this correct? For example,
```
A->B->C -+
       ^-----/
```
Cycle between B and C but if we build A first then `nodeAdded` will be true. Perhaps this needs to be `if len(nodesToAdd) != 0`?

request-manager/graph/grapher.go
```diff
@@ -287,30 +310,29 @@ func buildSeqGraph(seqSpec *spec.Sequence, idgen id.Generator) (seqGraph *Graph,
 		}
 	}
 
-	// Append sink node to ordering
 	seqGraph.Order = append(seqGraph.Order, seqGraph.Sink)
```
daniel-nichter, line 313: Another confusing part of the graph logic: in `newSeqGraph` the sink node is created, so why do we have to add it here? I really think `Graph` is leaking way too much logic.

request-manager/graph/grapher.go
```diff
@@ -287,30 +310,29 @@ func buildSeqGraph(seqSpec *spec.Sequence, idgen id.Generator) (seqGraph *Graph,
...
 
 	// Make sure we haven't created a deformed graph
 	// If we do, it's a bug in the code, not a problem with the specs
 	if !seqGraph.IsValidGraph() {
-		return nil, nil, fmt.Errorf("malformed graph created")
+		return nil, nil, fmt.Errorf("malformed sequence graph created")
```
daniel-nichter, line 318: We need to improve this by making `IsValidGraph()` return `error` that describes the problem. So, `return nil, nil, fmt.Errorf("sequence %s is not a valid directed acyclic graph: %s", err)`

request-manager/graph/grapher.go
```diff
@@ -287,30 +310,29 @@ func buildSeqGraph(seqSpec *spec.Sequence, idgen id.Generator) (seqGraph *Graph,
...
 }
 
-func newGraph(name string, idgen id.Generator) (*Graph, error) {
+func newSeqGraph(name string, idgen id.Generator) (*Graph, error) {
 	id, err := idgen.UID()
 	if err != nil {
 		return nil, err
 	}
-	source := newNoopNode(name+"_start", id)
+	source := newNoopSeqNode(name+"_start", id)
```
daniel-nichter, line 329: Would it break anything if we did s/start/begin?  Imho the standard pairs of terms are begin/end and start/stop (and source/sink in this context).

request-manager/graph/grapher.go
```diff
@@ -334,9 +356,11 @@ func newNoopNode(name, id string) *Node {
...
 
-// Get the minimal set of job args that the sequence starts with.
-// In the context of a wider request of which this sequence is a part, there may be more job
-// args available, but this sequence should not access them, so they are irrelevant for our purposes.
+// getAllSequenceArgs returns the minimal set of job args that the sequence
+// starts with.
+// In the context of a wider request of which this sequence is a part, there may
+// be more job args available, but this sequence should not access them, so they
+// are irrelevant for our purposes.
 func getAllSequenceArgs(seq *spec.Sequence) map[string]bool {
```
daniel-nichter, line 364: Little bit of a confusion here: comment says "minimal set" but the func name is "all".  The seems to return all args. Why, according to code comment, may there be more job args but this seq shouldn't access them?

request-manager/graph/graph.go
```diff
@@ -44,14 +45,14 @@ type Node struct {
 	SequenceRetryWait string                 // The time to sleep between sequence retries
```
daniel-nichter, line 45: Not sure the `Job job.Job` field is needed (above, not shown in GitHub diff). When we create a job node, we create the `job.Job`, then call its `Create` and `Serialize` methods, and then it's completely done; we don't have to keep a reference to it any longer. The important thing it does is mutate `jobArgs` and return a `[]byte` from its `Serialize` method.

request-manager/graph/graph.go
```diff
@@ -44,14 +45,14 @@ type Node struct {
...
+// IsValidGraph asserts that g is a valid graph by ensuring that
+// g is acyclic and connected (not fully connected), and that the edge
+// map matches the reverse edge map.
 func (g *Graph) IsValidGraph() bool {
 	return !g.hasCycles() && g.isConnected() && g.edgesMatchesRevEdges()
 }
 
-// Get out edges of node (node id --> Node)
+// GetNext returns the out edges of a node (node id --> Node)
 func (g *Graph) GetNext(n *Node) map[string]*Node {
```
daniel-nichter, line 56: No callers use the key, only the value. So the return can probably be `[]*Node`.

request-manager/graph/graph.go
```diff
@@ -44,14 +45,14 @@ type Node struct {
...
 	next := map[string]*Node{}
 	for _, nextId := range g.Edges[n.Id] {
```
daniel-nichter, line 58: Then you can also micro-optimize:
```go
next := make([]*Node, len(g.Edges[n.Id])
for i, id := range g.Edges[n.Id] {
    next[i] = g.Nodes[id]
}
```

request-manager/graph/graph.go
```diff
@@ -86,31 +88,31 @@ func (g *Graph) InsertComponentBetween(component *Graph, prev *Node, next *Node)
 		g.Nodes[k] = v
```
daniel-nichter, line 88: There's an error about (not shown by GitHub diff):
```go
return fmt.Errorf("Graph not valid!")
```
Let's improve that.

request-manager/graph/graph.go
```diff
@@ -155,51 +158,22 @@ func (g *Graph) PrintDot() {
...
+// isConnected returns true iff every node is reachable from the start node, and
+// every path terminates at the end node.
 func (g *Graph) isConnected() bool {
 	// Check forwards connectivity and backwards connectivity
 	return g.connectedToLastNodeDFS(g.Source) && g.connectedToFirstNodeDFS(g.Sink)
 }
 
-// Returns true if the last node in g is reachable from n
+// connectedToLastNodeDFS returns true iff the last node in g is reachable from n.
 func (g *Graph) connectedToLastNodeDFS(n *Node) bool {
```
daniel-nichter, line 177: Can drop the "DFS" part of all func names. In a case like this, best practice is not tightly coupling func to implementation/algorithm. In other cases, e.g. when using a Strategy pattern, then maybe yes it makes sense to have the various implementations named after themselves.

request-manager/graph/resolver.go
```diff
@@ -142,95 +144,65 @@ func (o *resolver) BuildJobChain(jobArgs map[string]interface{}) (*proto.JobChai
...
-			SequenceId:        node.SequenceId,
-			SequenceRetry:     node.SequenceRetry,
-			SequenceRetryWait: node.SequenceRetryWait,
-			State:             proto.STATE_PENDING,
-		}
-		jc.Jobs[jobId] = job
-	}
-
-	return jc, nil
+	return seqGraph, nil
```
daniel-nichter, line 147: It's not a sequence graph here, it's a request graph, so
```go
reqGraph, err := o.buildSequence(cfg)
...
return reqGraph, nil
```

request-manager/graph/resolver.go
```diff
@@ -142,95 +144,65 @@ func (o *resolver) BuildJobChain(jobArgs map[string]interface{}) (*proto.JobChai
...
 }
 
-// (Recursively) build a sequence.
+// buildSequence recursively builds a sequence. If a sequence graph node represents
+// a job, buildSequence creates the corresponding job. If a sequence graph node needs
+// to be expanded, i.e. it represents anything but a job, it is recursively expanded
+// with another call to buildSequence. buildSequence also chooses the correct path
+// for conditional nodes.
 func (o *resolver) buildSequence(cfg buildSequenceConfig) (*Graph, error) {
```
daniel-nichter, line 155: Good code comment. Nicely summarizes what this important func does.

request-manager/graph/resolver.go
```diff
@@ -142,95 +144,65 @@ func (o *resolver) BuildJobChain(jobArgs map[string]interface{}) (*proto.JobChai
...
-	for _, arg := range seq.Args.Static {
-		if _, ok := jobArgs[*arg.Name]; !ok {
-			jobArgs[*arg.Name] = *arg.Default
-		}
-	}
-
-	/* Build request graph based on sequence graph. */
-	seqGraph, ok := o.SequenceGraphs[sequenceName]
+	// Build request graph based on sequence graph.
+	seqGraph, ok := o.sequenceGraphs[sequenceName]
```
daniel-nichter, line 160: Is this copying the seq graph? I don't think so. If that's true, then the RG is clobbering the SG?

Also, var name should be `reqGraph` (`rg` is acceptable in this case because Go likes short var names and "rg" should be unique so it's easy to search/highlight).

daniel-nichter: Oh wait, no. Just below we do `reqGraph, err := o.newReqGraph(cfg.graphName, jobArgs)`.  Ok, so then this code should be moved closer to where it's first used. Then it makes it a little more clear to readers.

Also, code comments on this would really clarify it: we're fetching the SG only to walk its structure and build the RG in parallel. So no mutations to SG.

request-manager/graph/resolver.go
```diff
@@ -41,41 +40,44 @@ func NewResolverFactory(jf job.Factory, specs map[string]*spec.Sequence, seqGrap
...
+	jobFactory     job.Factory               // factory to create nodes' jobs
+	allSequences   map[string]*spec.Sequence // sequence name --> sequence spec
+	sequenceGraphs map[string]*Graph         // sequence name --> sequence graph
+	idGen          id.Generator              // generates UIDs for jobs
 }
 
+// RequestArgs takes user input args and returns them as a job args map, the form
+// expected by the Resolver later when building the request graph. Optional and
+// static args are filled in with their default values.
 func (o *resolver) RequestArgs(jobArgs map[string]interface{}) ([]proto.RequestArg, error) {
```
daniel-nichter, line 72: Time to let go the past: let's `r *resolver`.  (`o` is because this part of the code used to be called Omega.)

request-manager/graph/resolver.go
```diff
@@ -442,7 +413,8 @@ func chooseConditional(n *spec.Node, jobArgs map[string]interface{}) (string, er
 	return seqName, nil
 }
 
-// Split `each: list:element` values into the list (as a slice) and the element.
+// getIterators splits `each: list:element` values into the list (as a slice)
+// and the element.
 // Returns the element name, and the slice to iterate over when repeating nodes.
 // If there is no repetition required, then it will return an empty string, "",
 // and the singleton [""], to indicate that only one iteration is needed.
```
daniel-nichter, line 420: Maybe `Each(` is a better func name because the node spec field is called `each`. Then:
```
// Each returns list and element values for the "each:" node spec. For example, with node spec,
//
//   each:
//     - foos:foo
//     - bars:bar
//
// where jobArgs[foo]=[]string{a,b} and jobArgs[bar]=[]string{x,y}, the return values are,
//
//   []string{foo, bar}
//   [][]interface{}{
//       []interface{}{a, b}
//       []interface{}{x, y}
//   }
//
// The first return value is a list of the elements as their singular name. So with "foos:foo",
// the element name is "foo" and "foos" is the job arg providing the values in the second
// return value.
```

```

Comment (daniel-nichter) [reply request-manager/graph/resolver.go:420]: That ^ might be wrong. This func has always been confusing to me, i.e. what list and element refer to and how the return values are supposed to be used. Would be nice to make it more intuitive, which probably requires more specific names than "list" and "element" which are very generic.

File: request-manager/graph/resolver.go [hunk 30]
```diff
@@ -142,95 +144,65 @@ func (o *resolver) BuildJobChain(jobArgs map[string]interface{}) (*proto.JobChai
...
+			jobArgsCopy, err := remapNodeArgs(nodeSpec, jobArgs)
 			if err != nil {
 				return nil, err
 			}
 
-			// Add the element to the node args unless there is none for this node
+			// Add elements to the node args unless there are none
+			// for this node
 			for j, elt := range elements {
 				if elt != "" {
```
daniel-nichter, line 207: This block is unclear but very important. What's really happening here is: Given "each: foos:foo", we magically (good magic!) create jobArgs[foo]=jobArgs[foos][i]. This also means that `if elt != ""` is confusing because does it mean jobArgs[foos] can't have any empty string values? I.e. we shouldn't make "" a magical value here.

request-manager/graph/resolver.go
```diff
@@ -240,74 +212,72 @@ func (o *resolver) buildSequence(cfg buildSequenceConfig) (*Graph, error) {
 			}
 
 			// Add the "if" to the node args if it's present
```
daniel-nichter, line 214: This only needs to be done if `n.IsConditional()` (next block), right?

request-manager/graph/grapher.go
laureneliu: It's used in the same loop as the other vars created between this and the loop itself. Those vars are used only in the loop, though, whereas this and the ones above it are also returned, so this seems like optimal placement to me.

It's like a sandwich, if that makes sense. If not, ignore this line.

request-manager/graph/grapher.go
laureneliu: Sequence graph nodes need to be keyed on something. The alternative here is to just use the node name, which is unique within the sequence graph. I left it up to a UID generator because it seems better for the perhaps unlikely case where the sequence graph is extremely large, and the node names are inconvenient for whatever golang's default string hash function is (I can't seem to find the answer in the first three links of my first google search). For small graphs, maybe the overhead is greater than if I just keyed on node names, but it's not noticeably slow for our requests dir, so I left it.

request-manager/graph/grapher.go
laureneliu: The code used to save `len(nodesAdded)` for this purpose instead of using a flag. I realize that it was more efficient that way because we don't need to update `nodesAdded` on every iteration of this inner loop over nodesToAdd, but it was also somewhat less readable to me, since that was another var we needed to track from outside the outer loop.

request-manager/graph/grapher.go
daniel-nichter: Re previous comments and Slack message: I see now. So we can "solve" this with some illuminating code comments (either here are maybe on the inner loop), something to the effect of:
```go
// Each loop through the nodes, we should be able to add at least 1 node.
// If not, the there's a cycle (or perhaps a typo in deps: that wasn't caught
// earlier). For example, with A->B->C, 1st loop will add A since it has no
// deps, then it'll add B since it depends on A which has been built, then C.
// But if a cycle existed between B and C, the 2nd loop wouldn't be able to
// build B and nodeAdded would be false and trigger the error after the loop.
```

request-manager/graph/graph.go
laureneliu: Is it really okay for `GetNext` and `GetPrev` to have different return types like that?

request-manager/graph/resolver.go
laureneliu: `each:` is for sequence expansion, not conditional nodes. Conditional nodes are `if:` and `eq:`.

request-manager/graph/grapher.go
daniel-nichter: Ok, this is fine for now. My concern was that it's a bit confusing because at this point (i.e. in a SG), the concept of a node ID _slightly_ conflicts with the real node IDs which are part of request graphs. In an SG, nodes don't need a unique ID because they already have 2 unique identifies: name and spec order (not to be confused with resulting graph topological order). So in addition to slightly conflict with real node IDs, they're also not strictly needed here.

But we can leave it for now because this is better addressed later if/when we make a new `Graph` that does a better job of hiding all such graph details.

request-manager/graph/grapher.go
daniel-nichter: I've never heard of a sandwich in this sense, but sure. :-) Ok for now then.

request-manager/graph/grapher.go
daniel-nichter: We can resolve this. I think the other comment about detecting a cycle in A->B->C addressed this.

request-manager/graph/graph.go
daniel-nichter: No, that'd not be ideal, but I think (iirc the code) they can both return `[]*Node`?

request-manager/graph/resolver.go
daniel-nichter: Yes, but his isn't `each:`. What I'm suggesting is:
```go
			if nodeSpec.IsConditional() {
				// Node is a conditional: choose which path to
				// take, and recursively build the subgraph

				// Add the "if" to the node args if it's present
				if nodeSpec.If != nil {
					if ifArg, ok := jobArgs[*nodeSpec.If]; ok {
						jobArgsCopy[*nodeSpec.If] = ifArg
					}
				}
```
because `nodeSpec.If` is only valid if the node is a conditional.

request-manager/graph/grapher.go
```diff
@@ -361,26 +410,28 @@ func dependenciesSatisfied(inGraph map[string]bool, dependencies []string) bool
...
 
-	// Assert that the iterable variable is present.
+	// Assert that the iterable variable is present
 	for _, each := range n.Each {
 		if each == "" {
 			continue
 		}
 		if len(strings.Split(each, ":")) != 2 { // this is malformed input
-			return missing, fmt.Errorf("in node %s: malformed input to `each:`", n.Name)
+			return fmt.Errorf("in node %s: malformed input to `each:`", n.Name)
```
daniel-nichter, line 423: A better error message, but let's improve it more. In a software context, avoid "malformed" because it's imprecise and slightly anthropomorphic. For example, another system of ours calls TLS certs "fresh", but that's also imprecise for software. If it was a loaf of bread then sure: fresh or stale. But a TLS cert is valid or expired.

request-manager/graph/grapher.go
```diff
@@ -361,26 +410,28 @@ func dependenciesSatisfied(inGraph map[string]bool, dependencies []string) bool
...
 	}
 
-	// Assert that the conditional variable is present.
-	if n.If != nil {
+	// If this is a conditional node, assert that the "if" job arg is present
+	// in the job args map.
+	// Static checks assert that for conditional nodes, if != nil.
+	if n.IsConditional() {
 		if !jobArgs[*n.If] {
 			missing = append(missing, *n.If)
```
daniel-nichter, line 436: This is a different and more specific error than below, so let's return it as such. Something like,
`fmt.Errorf("in node %s: `if: %s`: job arg %s is not set", n.Name, n.If. nIf")`

request-manager/graph/grapher.go
```diff
@@ -393,5 +444,9 @@ func getMissingArgs(n *spec.Node, jobArgs map[string]bool) ([]string, error) {
 		}
 	}
 
-	return missing, nil
+	if len(missing) > 0 {
+		return fmt.Errorf("node %s missing job args: %s", n.Name, strings.Join(missing, ", "))
```
daniel-nichter, line 448: These jobs aren't "missing", they're not set. This slight change in wording makes a big difference because node specs have a `sets` field. So errors says "job args aren't set" clues the user into the problem: these args are in some node's `sets` list.

request-manager/graph/grapher.go
```diff
@@ -207,78 +238,93 @@ func buildSeqGraph(seqSpec *spec.Sequence, idgen id.Generator) (seqGraph *Graph,
...
+		// existed between B and C, the second loop wouldn't be able to
+		// build B and nodeAdded would be false and trigger the error
+		// after this loop.
+		for nodeName, node := range nodesToAdd {
 			nodeSpec := seqSpec.Nodes[nodeName]
-			if dependenciesSatisfied(nodesAdded, nodeSpec.Dependencies) {
-				// Dependencies for node have been satisfied; presumably, all input job args are
-				// present in job args map. If not, it's an error.
-				missingArgs, err := getMissingArgs(nodeSpec, jobArgs)
+			if !dependenciesSatisfied(nodesAdded, nodeSpec.Dependencies) {
```
daniel-nichter, line 277: I think s/dependenciesSatisfied/haveAllDeps/ makes the func more intention revealing. "Satisfied" is also anthropomorphic. Software has no emotions. :-) What we really mean here is: we can't build this node until we have (built) all its dependencies.

request-manager/graph/resolver.go
```diff
@@ -41,41 +40,44 @@ func NewResolverFactory(jf job.Factory, specs map[string]*spec.Sequence, seqGrap
...
 	return &resolver{
-		Request:           req,
-		JobFactory:        f.jf,
-		AllSequences:      f.seqs,
-		SequenceGraphs:    f.seqGraphs,
-		IdGen:             f.idf.Make(),
+		request:        req,
+		jobFactory:     f.jf,
+		allSequences:   f.seqs,
+		sequenceGraphs: f.seqGraphs,
```
daniel-nichter, line 46: For end to end consistency (which also makes it easy to search for vars that are the same in various components): s/allSequences/seqSpecs/ and s/sequenceGraphs/seqGraphs/ and also above in `resolverFactory` s/seqs/seqSpecs/. Then we've got two consistent vars throughout: seqSpecs and seqGraphs.

daniel-nichter: Similarly, we often use "seq" or "sequence" in var names. Let's use "seq" consistently because Go likes short var names. This is just for var names. In code comments, we can alternate between whichever reads best in context. I usually use full word first time it appears, then short form after.

request-manager/graph/resolver.go
```diff
@@ -121,193 +123,168 @@ func (o *resolver) RequestArgs(jobArgs map[string]interface{}) ([]proto.RequestA
...
-		elements, lists, err := o.getIterators(n, jobArgs)
+		nodeSpec := seqGraphNode.Spec
+
+		// 1. Do sequence expansion.
+		// Find out how many times this node has to be repeated.
+		// If this node has a `each:` field, then lists[i] and
+		// elements[i] describe the ith `list:element`. Specifically,
+		// lists[i] is the actual list value, not just the name of the
+		// job arg.
+		elements, lists, err := r.getIterators(nodeSpec, jobArgs)
```
daniel-nichter, line 190: 👍  Nice. I'm still not a big fan of list:element, but the code comment helps so much that I'm ok with it now. I have something to refer back to when I forget what list and element actually are/contain.

request-manager/graph/resolver.go
```diff
@@ -121,193 +123,168 @@ func (o *resolver) RequestArgs(jobArgs map[string]interface{}) ([]proto.RequestA
...
-			// Add the "if" to the node args if it's present
-			if n.If != nil {
-				if ifArg, ok := jobArgs[*n.If]; ok {
-					jobArgsCopy[*n.If] = ifArg
+			// If this is a conditional node, add the "if" job arg.
+			// Static checks asserted if != nil for conditional nodes.
+			if nodeSpec.IsConditional() {
+				if ifArg, ok := jobArgs[*nodeSpec.If]; ok {
+					jobArgsCopy[*nodeSpec.If] = ifArg
 				}
```
daniel-nichter, line 225: The inner `if {` block can go in the next block which starts with the same check: `if nodeSpec.IsConditional() {`.

request-manager/graph/resolver.go
```diff
@@ -121,193 +123,168 @@ func (o *resolver) RequestArgs(jobArgs map[string]interface{}) ([]proto.RequestA
...
-			// Add the new node to the map of completed components
-			components = append(components, reqSeqGraph)
+			components = append(components, reqSubgraph)
 
-			// If the node (or sequence) was determined to set any args
+			// If the node or sequence was determined to set any args
 			// copy them from jobArgsCopy into the main jobArgs
-			err = setNodeArgs(n, jobArgs, jobArgsCopy)
+			err = setNodeArgs(nodeSpec, jobArgs, jobArgsCopy)
 			if err != nil {
```
daniel-nichter, line 275: `if err := setNodeArgs(nodeSpec, jobArgs, jobArgsCopy); err != nil {`

Whenever the line is short enough, the `if err := func(); err != nil {` is a hallmark of idiomatic Go. Also, reusing `err` here is technically fine but imho it's a good practice to avoid for same reason we avoid global vars: the more locally scoped the better to avoid any chances of "cross-contamination".

request-manager/graph/resolver.go
```diff
@@ -121,193 +123,168 @@ func (o *resolver) RequestArgs(jobArgs map[string]interface{}) ([]proto.RequestA
...
+		elements, lists, err := r.getIterators(nodeSpec, jobArgs)
 		if err != nil {
-			return nil, fmt.Errorf("in seq %s, node %s: invalid 'each:' %s", sequenceName, n.Name, err)
+			return nil, fmt.Errorf("in seq %s, node %s: invalid 'each:' %s", sequenceName, nodeSpec.Name, err)
 		}
 
-		// All the graphs that make up this sequence graph node's subgraph
+		// All the graphs that make up this sequence graph node's subgraph,
+		// one for each time the node is repeated
 		components := []*Graph{}
```
daniel-nichter, line 197: Now that code is more clear, I think we'll enhance that clarify even further by getting ride of this very generic term that only appears here, once, and nowhere else. I think `seqExpansion` (or maybe `expandedSeq`) is better because that's it's primary purpose and that var name is immediately intention-revealing to FE who know basic Spin Cycle design elements.

And to really make the life of the FE easier, let's give an example:
```
// For example, with 3 sequences like S1->S2->S3, if we're currently
// in S1 (i.e. reqGraph=S1) and S2 is expanded ("each: foo:bar"),
// we'll build N-many parallel S2 and save them in seqExpansion.
// Then, after this `for` loop, we'll wrap all seqExpansion graphs
// in a new graph so they have a single source and sink node.
```

request-manager/graph/resolver.go
```diff
@@ -412,23 +390,24 @@ func (o *resolver) buildSequence(cfg buildSequenceConfig) (*Graph, error) {
 	return reqGraph, nil
```
daniel-nichter, line 390: Time to clean up part of code you didn't change but needs to be clarified, too. Above, the block that starts with,
```go
// Mark all vertices in sequence except start vertex start sequence id
```
It seems like we can and should be assigning sequence IDs while we're building each sequence? Also, the comment talks about, "A graph is built by constructing its inner most components, ..." which doesn't make sense in its own context, especially since the code block here isn't recursing or building anything.

request-manager/graph/resolver.go
```diff
@@ -556,22 +535,21 @@ func (o *resolver) buildSingleVertexGraph(nodeDef *spec.Node, jobArgs map[string
 	return g, nil
 }
 
-// NewReqGraph creates an "empty" graph. It contains two nodes: the "start" and "end" nodes. Both of these nodes
-// are no-op jobs
-func (o *resolver) newReqGraph(name string, jobArgs map[string]interface{}) (*Graph, error) {
+// newReqGraph creates an "empty" graph. It contains two nodes: the noop source and sink nodes.
+func (r *resolver) newReqGraph(name string, jobArgs map[string]interface{}) (*Graph, error) {
 	var err error
```
daniel-nichter, line 540: This var isn't really used until `err = setNodeArgs(&spec.NoopNode, jobArgs, jobArgsCopy)` (not shown in GitHub diff). Let's remove it and do,
```go
if err := setNodeArgs(&spec.NoopNode, jobArgs, jobArgsCopy); err != nil {
```

request-manager/graph/resolver.go
```diff
@@ -608,23 +585,28 @@ func (o *resolver) newNoopNode(name string, jobArgs map[string]interface{}) (*No
...
+			return nil, fmt.Errorf("Error making no-op job %s: %s", name, err)
 		}
 	}
 	err = rj.Create(jobArgs)
 	if err != nil {
-		return nil, fmt.Errorf("Error creating no-op node %s: %s", name, err)
+		return nil, fmt.Errorf("Error creating no-op job %s: %s", name, err)
+	}
+	bytes, err := rj.Serialize()
+	if err != nil {
```
daniel-nichter, line 596: Here's a great example of why I suggest `if err := func(); err != nil {` (i.e. block-scoping `err`): is `err` on this line from `bytes, err := rj.Serialize()` or `err = rj.Create(jobArgs)` just above? 🤔

request-manager/graph/resolver.go
```diff
@@ -650,43 +632,18 @@ func (o *resolver) newNode(j *spec.Node, jobArgs map[string]interface{}) (*Node,
...
+		return nil, fmt.Errorf("Error serializing '%s %s' job: %s", *j.NodeType, j.Name, err)
+	}
+
 	return &Node{
-		Job:       rj,
 		Name:      j.Name,
 		Id:        id,
+		Spec:      &spec.NoopNode, // on the next refactor, we shouldn't need to set this ourselves
```
daniel-nichter, line 643: Why is this `&spec.NoopNode` and not `j *spec.Node` ?

request-manager/graph/resolver.go
laureneliu: The code is slightly messier that way; right now, it does all the job arg stuff (starts with the `jobArgsCopy` stuff, ends with this block), then starts resolving the node. Moving the `if` block in with the next one mixes job arg stuff with resolving.

request-manager/graph/resolver.go
laureneliu: If we want to set sequence IDs while building it, we need to either 1) pass it to `buildSingleVertexGraph` and `newReqGraph`, and have the latter change its behavior based on use case (we're about to build a new request subgraph, or we're just using it to wrap expanded sequences); or 2) just set it all ourselves in `buildSequence` as we go.

I think it's better the way it is right now.

I'm not sure what's up with the comment--for one, it certainly seems to be setting `seqId` for the source and sink. For another, the comment initially says we don't set it for source, then says we don't set it for source or sink. And I have no idea what that TODO's about.

I remember being super confused by this comment when I was first reading the code, and just sort of hoping it'd make sense once I worked with the code more. It still makes no sense to me, so if you have any enlightening information, please share. Otherwise, I'll just update the comment to describe what I think is happening.

request-manager/graph/resolver.go
daniel-nichter: I see. Then I'd sell this as "separation of concerns" rather than clean/messy code. In essence we have,
```go
if X {
  // logic 1
}

if X {
  // logic 2
}
```
which raises a flag wrt another very valuable principle: DRY (Don't Repeat Yourself). An obvious duplication like that is messy code, imho, and a potential for bugs because engineers can forget to update the logic in both places (and it also raises questions about when, wrt `X`, we update logic 1, 2, or both).

But I see that you want to have all job arg logic together at top of loop, and all resolver logic together at bottom of loop. Separation of concerns is good, too. In other code, we might do that better by sticking all the job arg logic is a single func (or some other method of delineation). But this loop is small enough to keep it all in the loop. So to achieve virtual separation of concerns, i.e. signal to FE that we're going to duplicate code not by accident or oversight:

```go
		for i, _ := range lists[0] {
			// ---------------------------------------------------------------
			// Copy and prepare job args for this graph node

			...

			// ---------------------------------------------------------------
			// Resolve this graph node into a request subgarph

			var reqSubgraph *Graph
			if nodeSpec.IsConditional() {
```
Then the FE who's hopefully reading these should realize: job arg logic up here, and resolver logic down there, and maybe there's some overlap but that's ok because the separation is by design.

request-manager/graph/resolver.go
daniel-nichter: If it's not trivial to roll the seq ID into the nodes/req subgraphs while we're building them, then yes: let's just remove and rewrite the comment so it's correct and helpful.

daniel-nichter (approved):

PR #157, opened 2020-09-07
Add graph check for cyclical dependencies among sequences

laureneliu:


request-manager/graph/grapher.go
```diff
@@ -61,58 +61,103 @@ func (gr *Grapher) CheckSequences() (seqGraphs map[string]*Graph, seqErrors map[
...
-				missingSets[nodeName] = missing
+				// Do sets check.
+				missingSets := map[string][]string{} // node name -> list of missing `sets` args
+				for nodeName, nodeSpec := range seqSpec.Nodes {
+					if nodeSpec.IsJob() {
+						continue
+					}
+					// Compare declared `sets` with what the node
+					// actually sets
+					subseqs := getNodeSubsequences(nodeSpec, gr.sequenceSpecs)
```
felixplajer, line 133: This is pretty similar to what's happening around line 78. Is there a benefit to doing the 'sets' check here, or would it be possible to move it up there, so we don't have to repeat looping over all the nodes and finding subsequences?

request-manager/graph/grapher.go
```diff
@@ -61,58 +61,103 @@ func (gr *Grapher) CheckSequences() (seqGraphs map[string]*Graph, seqErrors map[
...
-				msg = append(msg, fmt.Sprintf("%s (failed to set %s)", nodeName, strings.Join(missing, ", ")))
-			}
-			multiple := ""
-			if len(missingSets) > 1 {
-				multiple = "s"
+		// If we were unable to check any sequences, there must be a
+		// cyclical dependency.
+		if !newSeqChecked {
+			for seqName, _ := range seqsToCheck {
+				seqErrors[seqName] = fmt.Errorf("part of cyclical dependency among sequences")
```
felixplajer, line 158: Does it matter if this ends up overwriting an existing error?

request-manager/graph/grapher.go
laureneliu: If we do it at line 78, the errors from sets checks can be somewhat inconsistent because the loop at 74 doesn't go through sequences in any particular order.

Suppose you have sequences A, B, and C s.t. A calls B calls C (C -> B -> A, since A depends on B depends on C). Furthermore, the node in B that calls C "sets" an arg that no node in C sets, i.e. there's a sets error in B. There are no other errors.

Suppose we do the sets check in the loop at line 74. If the loop at 74 goes through the sequences as A, B, C, then A will pass because `seqErrors` doesn't contain an entry for B. On the next iteration, we fail B, because it contains a sets error. However, if the loop at 74 goes through the sequences as with B before A, then we'll fail B for the sets error, then fail A, because B failed.

If we keep the sets check here (line 127 ig?) then we'll always check dependencies first, and propagate the error across sequences.

request-manager/graph/grapher.go
laureneliu: Actually, I hadn't noticed that it'd do that. As it is, this will only overwrite errors generated while building the sequence (with `buildSeqGraph`). I'm inclined to preserve the build error. What do you think?

felixplajer: I agree it'd be nice to keep the build error somehow, but I think all the sequences that are part of the cyclical dependency issue need to have this error, else it can get confusing.

eg. say there's a cycle of sequences A -> B -> C -> A, but C also had a build error. If we prioritize build > cycle error, than you'll see something like A = art of cyclical dependency, B = art of cyclical dependency, C = build error, which implies that C _isn't_ part of the cycle issue.

Is there any way to preserve both errors? If not, I think overwriting the build error is, while not great, the better option.

request-manager/graph/grapher.go
felixplajer: Ah, okay, let's keep it here then.

request-manager/graph/grapher.go
laureneliu: I could always concatenate the error strings, but that's not particularly pretty. Or possibly make `seqErrors` a `map[string][]string`, but that seems even worse.

In any case, I agree that prioritizing the cycle error makes sense.

felixplajer: Yeah, I guess let's just go with the cycle error for now.  `map[string][]string` would work, but it is... ugly, and it's not that bad to just paper over the other error while the cycle error exists.

felixplajer (approved):

PR #158, opened 2020-09-08
Replace the password in the MySQL DSN with asterisks before logging. Closes #154

### LINKED ISSUE #154: Config output should hide MySQL DSN on startup
Background:
- During startup, both request manager and job runner will dump their configs to stdout. For Request Manager, this includes the full MySQL DSN used to connect to the backend database, including username and plaintext password. 

Concern:
- Many deployments log stdout to a log aggregator that has expanded access beyond just the team with host level access. This poses a security risk by potentially exposing database credentials.

Suggested behavior:
- If possible, spincycle should hide this by default when `ENVIRONMENT=production` or allow a flag in the yaml or environment vars to hide it. Perhaps a `SPINCYCLE_DISPLAY_DSN` var that can be enabled if DB behavior is not correct and debugging is needed in production.

michaelfinch:


TeeKraken:
Thanks for this patch @michaelfinch, I imported it into our environment and confirmed that it starts up and does exactly what is needed!

felixp-square (approved):

PR #159, opened 2020-09-14
Improve comments+small optimization in grapher

laureneliu:


request-manager/graph/grapher.go
```diff
@@ -93,10 +101,6 @@ func (gr *Grapher) CheckSequences() (seqGraphs map[string]*Graph, seqErrors map[
 	// anymore, either because we've checked them all, or because impossible
 	// dependencies exist.
 	seqsChecked := map[string]bool{}
-	seqsToCheck := map[string]*spec.Sequence{}
-	for k, v := range gr.sequenceSpecs {
-		seqsToCheck[k] = v
```
felixplajer, line 98: tbh I'd actually slightly prefer to keep this here, instead of populating seqsToCheck in the earlier loop. Getting rid of looping over all the sequences one more time will only very minimally speed things up, and this loop makes it easy to see "ok, seqsToCheck just starts out with all the sequences" rather than scrolling up to find where it's set. That said it's not really a big deal to me so you can decide what you want to do.

felixplajer (approved):

PR #161, opened 2020-09-18
Improve linter CLI output

laureneliu:


linter/bin/main.go
```diff
@@ -3,7 +3,6 @@
 package main
```
daniel-nichter, line 3: `// Copyright 2020, Square, Inc.` above

request-manager/graph/grapher.go
```diff
@@ -149,19 +149,15 @@ func (gr *Grapher) CheckSequences() (seqGraphs map[string]*Graph, seqErrors map[
...
 
 		// If we were unable to check any sequences, there must be a cyclical dependency.
 		if !newSeqChecked {
 			for seqName, _ := range seqsToCheck {
-				// This overwrites a build error, if there was one. We want all
-				// sequences that were part of the cyclical dependency to have
-				// this error; otherwise, we imply that some sequence wasn't part
-				// of the cyclical dependency when it actually was.
-				seqErrors[seqName] = fmt.Errorf("part of cyclical dependency among sequences")
+				seqErrors[seqName] = append(seqErrors[seqName], fmt.Errorf("part of cyclical dependency among sequences"))
```
daniel-nichter, line 160: Can we report which other seq form the cycle? Would that be seqName in cycle with seqsToCheck[*] - seqName ?

request-manager/spec/errors.go
```diff
@@ -43,7 +39,7 @@ func (e InvalidValueError) Error() string {
 		multipleValues = "s"
 	}
 
-	return fmt.Sprintf("%s: invalid value%s %s in field '%s', expected %s",
+	return fmt.Sprintf("%sinvalid value%s %s in field '%s', expected %s",
```
daniel-nichter, line 42: Typo: no space between "%s" and "invalid"

request-manager/spec/errors.go
```diff
@@ -75,15 +67,14 @@ func (e MissingValueError) Error() string {
 		explanation = fmt.Sprintf(": %s", e.Explanation)
 	}
 
-	return fmt.Sprintf("%s: field '%s' missing%s", loc, e.Field, explanation)
+	return fmt.Sprintf("%sfield '%s' missing%s", loc, e.Field, explanation)
```
daniel-nichter, line 70: Typo: missing space

request-manager/spec/parser.go
```diff
@@ -12,39 +12,33 @@ import (
 )
 
 // Parse a single request (YAML) file.
-// 'logFunc' is a Printf-like function used to log warning(s) should they occur.
-// Errors are returned, not logged.
-func ParseSpec(specFile string, logFunc func(string, ...interface{})) (Specs, error) {
-	var spec Specs
-
+func ParseSpec(specFile string) (spec Specs, err error, warning error) {
```
daniel-nichter, line 15: In Go, `error` is pretty much always the last return arg:
```go
ParseSpec(specFile string) (spec Specs, warning error, err error)
```
Normally, don't use named args unless there's an explicit reason for doing so, but this maybe a case with a reason: two return `error` is confusing unless they're named.

request-manager/spec/errors.go
laureneliu: These are deliberate; either %s is the empty string, in which case we don't want a leading space, or it's "node: ", which comes with a space.

PR #167, opened 2020-09-23
Small improvement in linter CLI

laureneliu:


linter/linter.go
```diff
@@ -147,18 +151,19 @@ func Run() bool {
...
 		}
-		return false
-	}
-
-	// 5. No errors occurred: print warnings, then exit.
-	if seqResults.AnyWarning {
+		if errorPrinted {
+			return false
+		}
+	} else if seqResults.AnyWarning {
```
felixplajer, line 165: I think this should be just if, not else if. Else, if there are errors but not for the specified `--sequences`, you won't print any warnings for those sequences even if they exist.

linter/linter.go
```diff
@@ -205,9 +212,9 @@ func (linter *Linter) fmtList(header string, list []error) string {
 	return str
 }
 
-func (linter *Linter) printCheckResult(header string, result *spec.CheckResult) {
+func (linter *Linter) printCheckResult(header string, result *spec.CheckResult) bool {
```
felixplajer, line 215: I'd name the return value to make it clear what it's indicating — `printed` or `didPrint` or something like that.

linter/linter.go
```diff
@@ -168,6 +173,8 @@ func Run() bool {
 		}
 	}
 
+	// 5. No errors occurred, and we've printed all warnings. Decide whether
+	// we need to print anything else, then exit.
 	if linter.anyWarning && !linter.Warnings {
```
felixplajer, line 178: Minor thing, but this will say warnings were suppressed even if there were only warnings for sequences not specified in `--sequences`.

felixplajer: Oh, uh, also, `linter.anyWarning` doesn't seem to take into account `seqResults.AnyWarning`?

linter/linter.go
```diff
@@ -147,18 +151,23 @@ func Run() bool {
...
+			errorPrinted = linter.printCheckResult(header, seqResults.Results[seq]) || errorPrinted
+		}
+		if errorPrinted {
+			return false
 		}
-		return false
 	}
 
-	// 5. No errors occurred: print warnings, then exit.
+	// 5. No errors occurred, and we've printed all warnings. Decide whether
```
felixplajer, line 167: "we've printed all warnings"  but there's still warnings to print

laureneliu: -_-

felixplajer (approved):

PR #169, opened 2020-11-19
Update depth-first search algorithms

michaelfinch:
Update depth-first search algorithms in the hasCycles() and isConnected() grapher methods to keep track of visited nodes and skip them on future iterations. Previously we were iterating through every possible path in the graph, which is unnecessary.

This change greatly reduces the creation time of a graph. For example, a moderately complex graph with a few hundred jobs that used to take 4 minutes to create now takes less than 1 second.

### DIFF
--- a/request-manager/graph/graph.go
+++ b/request-manager/graph/graph.go
@@ -173,23 +173,31 @@ func (g *Graph) PrintDot() {
 
 // hasCycles returns true iff the graph has at least one cycle in it.
 func (g *Graph) hasCycles() bool {
-	seen := map[string]*Node{g.Source.Id: g.Source}
-	return g.hasCyclesHelper(seen, g.Source)
+	inPath := map[string]*Node{g.Source.Id: g.Source}
+	visited := map[string]struct{}{}
+	return g.hasCyclesHelper(inPath, visited, g.Source)
 }
 
 // isConnected returns true iff every node is reachable from the start node, and
 // every path terminates at the end node.
 func (g *Graph) isConnected() bool {
+	lvisited := map[string]struct{}{}
+	fvisited := map[string]struct{}{}
 	// Check forwards connectivity and backwards connectivity
-	return g.connectedToLastNode(g.Source) && g.connectedToFirstNode(g.Sink)
+	return g.connectedToLastNode(g.Source, lvisited) && g.connectedToFirstNode(g.Sink, fvisited)
 }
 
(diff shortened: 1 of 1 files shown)


request-manager/graph/graph.go
```diff
@@ -173,57 +173,72 @@ func (g *Graph) PrintDot() {
 
 // hasCycles returns true iff the graph has at least one cycle in it.
 func (g *Graph) hasCycles() bool {
-	seen := map[string]*Node{g.Source.Id: g.Source}
-	return g.hasCyclesHelper(seen, g.Source)
+	inPath := map[string]*Node{g.Source.Id: g.Source}
+	visited := map[string]*Node{}
```
felixp-square, line 177: You could make all the `visited` maps `map[string]struct{}` to save a bit of space (since the stored *Node is never actually retrieved).

michaelfinch: Good point, I will fix that

request-manager/graph/graph.go
```diff
@@ -250,26 +265,37 @@ func (g *Graph) edgesMatchesRevEdges() bool {
...
+// Determines if a graph has cycles, using dfs. inPath is a list of nodes
+// that are in the current traversal path. visited is a list of nodes that
+// are guaranteed to not be involved in any cycles.
+// precondition: start node is already in inPath list
+func (g *Graph) hasCyclesHelper(inPath map[string]*Node, visited map[string]struct{}, start *Node) bool {
 	for _, next := range g.GetNext(start) {
+		// If we've already determined the next node is not involved in
+		// any cycles, return false
+		if _, ok := visited[next.Id]; ok {
+			return false
```
felixp-square, line 277: This needs to be just a `continue` instead of a return.

See this [example graph](https://docs.google.com/drawings/d/15R2qd9sbC1U7e_MBrE2qWj5LLYP48KkbPih-lydh_vU/edit?usp=sharing). Suppose `hasCyclesHelper` is first called on node A — it will visit node C, add that to the `visited` map since there are no cycles, and then return. Later `hasCyclesHelper` is called on node B. It loops through the nodes "after" B (i.e. C + D) — let's say C is done first. Since C has already been visited, it returns here saying "there are no cycles", missing the cycle between nodes D + E.

michaelfinch: Great catch. Fixed

felixp-square (approved):

PR #176, opened 2022-02-14
Add Support for Jobs Ending in "UNKNOWN" State

Areson:
Adding support for jobs that end in an "UKNOWN" state. Jobs in unknown states are treated as a failure state instead of causing a panic. If possible the sequence should be retried as with a normal failure and if that is not possible or it is not able to be retried then the request should end in failure and not a panic.

### DIFF
--- a/job-runner/chain/chain.go
+++ b/job-runner/chain/chain.go
@@ -124,7 +124,7 @@ func (c *Chain) IsDoneRunning() (done bool, complete bool) {
 			}
 			// This job is pending but not runnable which means a previous job
 			// failed.
-		case proto.STATE_FAIL:
+		case proto.STATE_FAIL, proto.STATE_UNKNOWN:
 			// If sequence can retry, then chain isn't done or complete,
 			if c.canRetrySequence(job.Id) {
 				return false, false
@@ -150,7 +150,7 @@ func (c *Chain) FailedJobs() uint {
 	defer c.jobsMux.RUnlock()
 	n := uint(0)
 	for _, job := range c.jobChain.Jobs {
-		if job.State == proto.STATE_FAIL {
+		if job.State == proto.STATE_FAIL || job.State == proto.STATE_UNKNOWN {
 			n++
 		}
 	}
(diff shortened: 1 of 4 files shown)


felixp-square:
At line 151 in chain.go, does the unknown state also need to be incorporated into the FailedJobs() func? I'm thinking yes because that func is used in the stopped/suspedened reapers to tell if a chain is failed or not.

job-runner/chain/chain.go
```diff
@@ -124,6 +124,9 @@ func (c *Chain) IsDoneRunning() (done bool, complete bool) {
 			}
 			// This job is pending but not runnable which means a previous job
 			// failed.
+		case proto.STATE_UNKNOWN:
+			// Treat unknown as a failure to retry, is possible.
+			fallthrough
 		case proto.STATE_FAIL:
```
felixp-square, line 130: you can have multiple items in a case statement, so I'd switch the fallthrough for `case proto.STATE_FAIL, proto.STATE_UNKNOWN:` just because using fallthrough is pretty unusual in Go

job-runner/chain/reaper_test.go
```diff
@@ -1037,6 +1357,136 @@ func TestSuspendedReaper(t *testing.T) {
...
+			"job1": {"job2", "job3"},
+			"job2": {"job4", "job5", "job6"},
+			"job3": {"job4"},
+		},
+	}
+
+	// Make job6 a separate sequence
+	job := jc.Jobs["job6"]
+	job.SequenceId = "job6"
+	job.SequenceRetry = 1
```
felixp-square, line 1386: eg. related to my comment on FailedJobs, if you did a test like this but without giving job6 any retries, what you'd expect is for the request to fail, but as is I don't think it would.

job-runner/chain/chain.go
Areson: Done.

felixp-square (approved):
lgtm. I don't know why the builds didn't kick off, I'll have to poke around tomorrow and see what's up (or you can if you want).

Areson:
I poked around a little but but I didn't find anything obvious about where the checks are setup.

felixp-square:
I'm seeing if I can get the testing set up using Github Actions which seems to be the way forward, but in the meantime if you've tested it locally you can go ahead and merge the PR anyways.

PR #177, opened 2022-02-15
Add Options to output from "start"

Areson:
Updated the full command output by `start` to include relevant options that can affect the way the command runs. This includes: env, addr, and config. Adding basic escaping in the event the options include spaces.

### DIFF
--- a/spinc/cmd/start.go
+++ b/spinc/cmd/start.go
@@ -4,6 +4,7 @@ package cmd
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/square/spincycle/v2/proto"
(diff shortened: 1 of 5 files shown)


felixp-square:
2 things:

- There's also the part where it prints out "spinc status <id>" at line 191 which needs the options added to it.
- Somehow I didn't put two and two together that our build of spinc uses hooks when you mentioned on the ticket that flags set by hooks would show up as well. It does look somewhat cluttered always printing out the addr and env... What do you think about adding another field to the spinc Context that saves just the options set via the command line, which could then be used when figuring out the options to include in start?

Areson:
I'll mull around with the context idea. One might counter that doing so is hiding what spinc is really doing in terms of the command that is being invoked. Not that hiding the items set by the hook really changes the outcome.

felixp-square:
Hmm ok, I see the issue. One could work around that by changing the Options struct to separate the env var values and command line values, like
```
type Options struct {
	Addr    string `yaml:"addr"`
	EnvAddr string `arg:"env:SPINC_ADDR"`
	...
}
```
and then afterwards using that to figure out the final addr, etc that we want. ...It's not the most beautiful solution.

I do think what I'd like best as a user is if spinc only prints out flags that have been specified on the command line and not in env vars / config / hooks. That gives the minimal amount of extra flags to have the printed command hit the same addr/env as the original. I think it's reasonable to assume that the commands spinc prints out are meant to be run in the same environment as the original command, so the env vars + config files won't change.

felixp-square:
A typical way to get around that would be to change the Options fields to pointers, so you can differentiate the default value vs no value provided. However I don't know if the parsing package we're using works with pointers.

Areson:
Yup. Found that looking over the documentation for the arg parsing package. Added a separate struct for parsing the user options and merged that back into the context after the hooks so we have a clean set of options that can only come from the user.

spinc/cmd/start.go
```diff
@@ -188,16 +189,51 @@ func (c *Start) Run() error {
...
+		fullCmd = "--env " + escapeArg(c.ctx.Options.Env) + " "
+	}
+
+	if c.ctx.UserOptions.Addr != "" {
+		fullCmd += "--addr " + escapeArg(c.ctx.Options.Addr) + " "
+	}
+
+	if c.ctx.UserOptions.Config != "" {
+		fullCmd += "--config " + escapeArg(c.ctx.Options.Config) + " "
+	}
```
felixp-square, line 234: Since this is duplicating what statusUserOptions does, you could just call that func

felixp-square (approved):

PR #190, opened 2023-05-11
Update find endpoint to allow filtering by request args

jplulu:
Also update spinc cli find to work with the new filter.

- [x] Test in staging

### DIFF
--- a/request-manager/request/manager.go
+++ b/request-manager/request/manager.go
@@ -644,42 +644,6 @@ func (m *manager) Find(filter proto.RequestFilter) ([]proto.Request, error) {
 		}
 
 		requests = append(requests, req)
-
-		// if len(filter.Args) > 0 {
-		// 	if filter.Limit != 0 && uint(len(requests)) >= filter.Limit {
-		// 		break
-		// 	}
-		// 	if len(reqArgsBytes) > 0 {
-		// 		var reqArgs []proto.RequestArg
-		// 		if err := json.Unmarshal(reqArgsBytes, &reqArgs); err != nil {
-		// 			return []proto.Request{}, fmt.Errorf("Error unmarshaling request args: %s", err)
-		// 		}
-
-		// 		// build a map of argName -> argValue
-		// 		argsMap := make(map[string]interface{})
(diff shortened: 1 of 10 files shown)


request-manager/request/manager.go
```diff
@@ -540,39 +540,46 @@ func (m *manager) GetWithJC(requestId string) (proto.Request, error) {
...
 		for _, state := range filter.States {
 			values = append(values, state)
 		}
 	}
+	if len(filter.Args) != 0 {
+		for arg, val := range filter.Args {
+			fields = append(fields, "ra.create_request LIKE CONCAT('%\"', ?, '\":\"', ?, '\"%')")
+			values = append(values, arg, val)
+		}
+	}
```
felixp-square, line 568: Could you add a test for the args filter to [TestFind](https://github.com/square/spincycle/blob/cf48fe75816ae2e0932844dd8069e1757329d2d9/request-manager/request/manager_test.go#L662)?

spinc/cmd/find.go
```diff
@@ -255,6 +268,7 @@ Filters:
   type        type of request to return
   states      comma-separated list of request states to include
   user        return only requests made by this user
+  requestArgs return requests made with specific args (format: arg1=value1,arg2=value2)
```
felixp-square, line 271: `requestArgs` -> `args` I think, cause you're grabbing "args" from the map at line 108

felixp-square (commented):
couple of minor comments but overall lgtm

request-manager/request/manager.go
jplulu: I was looking to add a test for this but it seems that the existing test framework does not use/populate the `request_archives` table. So args will always be empty.

spinc/cmd/find.go
jplulu: done

request-manager/request/manager.go
jplulu: I'm not sure how to update it to populate `request_archives` table

felixp-square (approved):

PR #192, opened 2023-08-09
Update `Arg` for specs to use `interface{}` instead of `string`

Areson:
Currently the `Arg` struct for spec arguments parses argument values in the spec as `string` types. This causes some pain for jobs that are running using values directly from the spec as they cannot be defined using more meaningful types (e.g. `int` or `bool`) and instead must expect values to be strings. Changing this to use `interface{}` allows values to be parsed by the yaml parser into primitives based on the content of the yaml file rather than forcing everything to be strings.


daniel-nichter (approved):
Talked with @Areson and this should be ok in Spin Cycle itself since the Job method is already `Create(jobArgs map[string]interface{})`. However, it could be a breaking change in users' jobs if a job presumes `string` types. Now jobs need to "dereference" the interface{} to its real underlying type. So I suggest a new minor version for this change.

